PR #15, opened 2016-09-27
Update README.md

Robinlovelace:


edzer:
I don't see any benefits in doing it this way; we want to build on released versions of udunits anyway.

Robinlovelace:
Fair enough - just put that because the tarball from the website didn't download when I was in Albacete. May have been the internet connection there - more for my own benefit.

PR #57, opened 2017-09-20
Add basic handling of logical NA values.

gmp216:
This PR simplifies the handling of NA values, allowing a user to write set_units(NA,"m/s") without first converting NA values (which default to logical) to numeric.


R/conversion.R
```diff
@@ -43,6 +43,17 @@ convert <- function(value, from, to) {
   x
 }
 
+`units<-.logical` <- function(x, value) {
+  if(is.na(x)) {
```
edzer: if (all(is.na(x)) {

R/conversion.R
```diff
@@ -43,6 +43,17 @@ convert <- function(value, from, to) {
...
+    c <- match.call()
+    c[1] <- call('units<-.numeric')
+    c[['x']] <- as.numeric(x)
+    eval(c)
```
edzer: I'd prefer to use .NextMethod() here, as in
```
x <- as.numeric(x)
NextMethod()
```

R/conversion.R
```diff
@@ -43,6 +43,17 @@ convert <- function(value, from, to) {
...
+  } else {
+    stop("Error in units<-(x,...) : x must be class units, numeric or NA")
+  }
```
edzer: suggest: `stop("x must be numeric, non-NA logical not supported")`

R/conversion.R
```diff
@@ -128,6 +139,17 @@ set_units.numeric = function(x, value = units::unitless, ...) {
   }
 }
 
+set_units.logical = function(x, value = units::unitless, ...) {
+  if (is.na(x)) {
```
edzer: (same comments)

R/conversion.R
gmp216: NextMethod() does not appear to work here because the logical values provided only have a single type and not a hierarchy. There may be a way to make it work but it could end up more complex than my version. Interested to see any other ways to do this.

edzer: Thanks for trying, let's leave it like this.

edzer:
Great, thanks!

PR #59, opened 2017-09-20
Guarded assignment operator

gmp216:
It is often useful when working with units to ensure that an assignment doesn't change the units of a target variable. In R this requires the definition of a new assignment operator. This could be helpful for debugging purposes or to ensure that output variables are in desired units for plotting or analysis.

This pull request defines a %<% operator implementing this functionality (open to suggestions for a better name). The operator will convert units that are commensurate or throw a useful error message if not.

example:
>  x <- set_units(1, mi/h)
>  y <- set_units(100,km/h)
>  x %<% y
>  x
62.13712 mi/h


edzer:
What is wrong with
```r
> set_units(y, units(x))
62.13712 mi/h
```
?

gmp216:
Nothing is wrong with that approach, just that it might be helpful to have some syntactic sugar so users can write equations in a more natural way while preserving strict unit checking and translation. I realize now %<% does not really accomplish that because of operator precedence (x %<% y + z will parse as (x %<% y) + z, which does not work for my purposes), so we would actually have to override <- itself to make this work which I am sure nobody wants to do. Note that set_units(...) has this issue too and can result in some surprising things:

> y <- set_units(100,km/h)
> x <- y + y %>% set_units(units(x)) # BUG
> x
200 km/h

Of course you need to say "x <- (y + y) %>% set_units(units(x))" or "set_units(y + y, units(x))" to avoid this but this looks like it might end up happening a lot.

An alternative approach could look something like this:

`value<-.units` <- function(x, value) {
   set_units(value,units(x))
}

>  value(x) <- y + y
> x
124.2742 mi/h

This could be used during development and debugging and then removed selectively if it is no longer necessary and performance is an issue. If this syntax can help users avoid bugs in using the units package itself, it could be worthwhile. But this feature is more of a possible "nice-to-have", and since it is clearly not ready to merge right now I'm happy to have this closed or moved to "issues" if you think it might be helpful for discussion with other end users.

PR #61, opened 2017-09-28
Implement units(), units()<- and set_units() for data frames.

gmp216:
I thought it would be useful to have a more convenient way to set units on dataframes. This PR implements units()<- and set_units() S3 methods for data frames, which accept either single values of class 'units' or lists of units objects. The lists can be either unnamed lists of the same length as the number of columns of the target data frame, or named lists whose names match some or all of the column names in the target dataframe.

For example:

# set all columns that are convertible to 'units' (ignores columns that are factors, characters, etc.)
> units(df) <- with(ud_units, m)
> df <- set_units(df, m)

# set named columns (a and b must be names in df):
> units(df) <- with(ud_units, list(a=m, b=mi/hr))
> df <- set_units(df, list(a=m, b=mi/hr))

# set units on all columns in a two-column data frame:
> units(df) <- with(ud_units, list(m, mi/hr))
> df <- set_units(df, list(m, mi/hr))

This also works on dataframe slices:

> df[,c('a','b')] <- with(ud_units, list(m, mi/hr))

The PR also implements units.data.frame() to retrieve a named list of units objects corresponding to the units in each column of the dataframe that are of type 'units'. This allows you to do things like this:

> units(df2) <- units(df1)


edzer:
I don't think that this is a good idea, after all.

Issue #60, opened 2017-09-22
make_units2() and .udunits_symbols_info()


t-kalinowski:
Hello,

As part of a project I recently built something ontop of the `units` package, and in the process I made some helper functions that might be beneficial to others.

### make_unit2()
The first is `make_unit2()`. (There is probably a better name for it, but this was the best I could come up with after a few minutes of thought).

This was to address some issues I had with `units::make_unit()`. The first is that `make_unit` does not parse strings correctly as expressions into denominator and numerator fields of the “symbolic_units” object.
```
format(units(make_unit("ug/l")))
##   numerator denominator 
##      "ug/l"          ""
format(units(make_unit2("ug/l")))
##   numerator denominator 
##        "ug"         "l"
```
This had downstream effects in that it a) interfered with unit simplification, and b) units were printed in parenthesis by default ("
```
format(make_unit("ug/l"))
## [1] "1 (ug/l)"
format(make_unit2("ug/l"))
## [1] "1 ug/l"
```
Finally, another issue was that make_unit would create units that were not valid or did not return true for `udunits2::ud.is.parseable()`. I think that flexibility is important to have, but I needed a version that would throw an error instead of create an invalid or unrecognized unit.
```
make_unit("foo")
## 1 foo
try(make_unit2("foo"))
```
`make_unit2()` is somewhat fancy in that it works with both strings or expressions. Here are some extended examples I included in the documentation:
```
# different ways to specify units
 # all 3 should be identical
 make_unit2(ug/l)
## 1 ug/l
 make_unit2("ug/l")
## 1 ug/l
 string <- "ug/l"
 make_unit2(string)
## 1 ug/l
 # valid unit names not found in units::ud_units also parse
 make_unit2(ug/gallon)
## 1 ug/gallon
 make_unit2("ug/gallon")
## 1 ug/gallon
 string <- "ug/gallon"
 make_unit2(string)
## 1 ug/gallon
 # the normal evaluation of the supplied argument is attempted first before
 # inspecting supplied expression. The expression is only inspected if it does
 # not resolve to a character vector"
 ug <- "kilogram"
 make_unit2(ug)
## 1 kilogram
 # note that even if one of the symbols is bound, NSE is done on the whole
 # expression or not at all
 make_unit2(ug/l)
## 1 ug/l
 # to avoid expression parsing, use argument .SEonly (e.g., defensive programmig
 # inside a function)
 # make_unit2(ug/l, .SEonly = TRUE) # ERROR
[...]

PR #64, opened 2017-10-13
overhaul

t-kalinowski:
add make_units() parse_units().
refactor set_units(), no longer a generic
add symbolic_unit(), replaces make_unit()
depracate make_unit() and parse_unit() (singular 'unit' functions to reduce confusion)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,8 +2,9 @@
 
 * Unit creation has been significantly refactored. `units<-` now accepts strings
 or quoted language objects on the right hand side, powered by new S3 methods for
-`as_units`. New user facing function `make_units()` (plural s) is also provided.
-See `?as_units` for details. @t-kalinowski
+`as_units`. All valid unit symbols and unit names recognized by package 'udunits2' are 
+now accepted. New user facing function `make_units()` (plural s) is also
+provided. See `?as_units` for details. @t-kalinowski
 
 * new functions `valid_udunits()` and `valid_udunits_prefixes()` generate tidy
 dataframes listing all the valid unit names, symbols, and prefixes recognized by
--- a/R/arith.R
+++ b/R/arith.R
@@ -155,8 +155,8 @@ Ops.units <- function(e1, e2) {
 #' a = set_units(1:5, m)
 #' a %*% a
 #' a %*% t(a)
-#' a %*% set_units(1:5, set_units(1))
-#' set_units(1:5, set_units(1)) %*% a
+#' a %*% set_units(1:5, 1)
+#' set_units(1:5, 1) %*% a
 `%*%.units` = function(x, y) {
 	ret = `%*%.default`(unclass(x), unclass(y))
 	units(ret) = .multiply_symbolic_units(1, units(x), units(y))
--- a/R/make_units.R
+++ b/R/make_units.R
@@ -333,10 +333,10 @@ See ?as_units for usage examples.")
   names(vars) <- vars
   tmp_env <- lapply(vars, symbolic_unit, check_is_valid = FALSE)
   
-  unit <- try(eval(x, tmp_env, units_eval_env), silent = TRUE)
-  if(inherits(unit, "try-error"))
-    stop(paste0(unit, "\n", 
-      "Did you try to supply a value in a context where a bare expression was expected?"))
+  unit <- tryCatch( eval(x, tmp_env, units_eval_env),
+    error = function(e) stop( paste0( conditionMessage(e), "\n",
+          "Did you try to supply a value in a context where a bare expression was expected?"
+        ), call. = FALSE ))
   
   if(as.numeric(unit) %not_in% c(1, 0)) # 0 if log() used. 
     stop(call. = FALSE,
--- a/R/make_units.R
+++ b/R/make_units.R
@@ -314,9 +314,14 @@ as_units.call <- function(x, check_is_valid = TRUE, ...) {
   
   stopifnot(is.language(x))
   
-  vars <- all.vars(x)
-  if(!length(vars)) 
+  if(missing(x))
     return(structure(1, units = unitless, class = "units"))
+  
+  vars <- all.vars(x)
+  if(!length(vars))
+    stop(call. = FALSE,
+"No symbols found. Please supply bare expressions with this approach.
+See ?as_units for alternative ways of creating units (e.g., from strings)")
   
   if (check_is_valid) {
     valid <- vapply(vars, is_valid_unit_symbol, logical(1L))
@@ -327,7 +332,10 @@ as_units.call <- function(x, check_is_valid = TRUE, ...) {
   names(vars) <- vars
   tmp_env <- lapply(vars, symbolic_unit, check_is_valid = FALSE)
   
-  unit <- eval(x, tmp_env, units_eval_env)
+  unit <- try(eval(x, tmp_env, units_eval_env), silent = TRUE)
+  if(inherits(unit, "try-error"))
+    stop(paste0(unit, "\n", 
+      "Did you try to supply a value in a context where a bare expression was expected?"))
   
   if(as.numeric(unit) %not_in% c(1, 0)) # 0 if log() used. 
     warning(call. = FALSE,
--- a/R/make_units.R
+++ b/R/make_units.R
@@ -151,7 +151,7 @@ are_exponents_implicit <- function(s) {
   s <- trimws(s)
   has <- function(chr, regex = FALSE) 
     grepl(chr, s, fixed = !regex, perl = regex)
-  !has("^") && !has("*") && !has("/") && has("\\s|\\d$", regex = TRUE)
+  !has("^") && !has("*") && !has("/") && has("\\s|\\D.*\\d$", regex = TRUE)
 }
 
 is_udunits_time <- function(s) {
@@ -312,16 +312,17 @@ units_eval_env$lb <- function(x) base::log(x, base = 2)
 #' @seealso \code{\link{valid_udunits}}
 as_units.call <- function(x, check_is_valid = TRUE, ...) {
   
-  stopifnot(is.language(x))
-  
-  if(missing(x) || tryCatch(eval(x, baseenv()) == 1, error = function(e) FALSE))
+  if(missing(x) || identical(x, quote(expr =)) || 
+     identical(x, 1) || identical(x, 1L))
     return(structure(1, units = unitless, class = "units"))
+  
+  stopifnot(is.language(x))
   
   vars <- all.vars(x)
   if(!length(vars))
     stop(call. = FALSE,
 "No symbols found. Please supply bare expressions with this approach.
-See ?as_units for alternative ways of creating units (e.g., from strings)")
+See ?as_units for usage examples.")
   
   if (check_is_valid) {
     valid <- vapply(vars, is_valid_unit_symbol, logical(1L))
--- a/R/make_units.R
+++ b/R/make_units.R
@@ -314,7 +314,7 @@ as_units.call <- function(x, check_is_valid = TRUE, ...) {
   
   stopifnot(is.language(x))
   
-  if(missing(x))
+  if(missing(x) || tryCatch(eval(x, baseenv()) == 1, error = function(e) FALSE))
     return(structure(1, units = unitless, class = "units"))
   
   vars <- all.vars(x)
--- a/R/make_units.R
+++ b/R/make_units.R
@@ -339,9 +339,8 @@ See ?as_units for usage examples.")
       "Did you try to supply a value in a context where a bare expression was expected?"))
   
   if(as.numeric(unit) %not_in% c(1, 0)) # 0 if log() used. 
-    warning(call. = FALSE,
-"In ", sQuote(deparse(x)), " the numeric multiplier ", sQuote(as.numeric(unit)), " was discarded. 
-The returned unit object was coerced to a value of 1.
+    stop(call. = FALSE,
+"In ", sQuote(deparse(x)), " the numeric multiplier ", sQuote(as.numeric(unit)), " is invalid. 
 Use `install_conversion_constant()` to define a new unit that is a multiple of another unit.")
   
   structure(1, units = units(unit), class = "units")
--- a/R/set_units.R
+++ b/R/set_units.R
@@ -32,9 +32,8 @@ set_units <- function(x, value, ...,
   }
   
   if (is.null(value))
-    drop_units(x)
-  else {
-    units(x) <- as_units(value, ...)
-    x
-  }
+    return(drop_units(x))
+  
+  units(x) <- as_units(value, ...)
+  x
 }
--- a/R/set_units.R
+++ b/R/set_units.R
@@ -9,17 +9,19 @@
 #'
 #' @param value a \code{units} object, or something coercible to one with
 #'   \code{as_units}. Depending on \code{mode}, the unit is constructed from the
-#'   supplied bare expression or from the supplied value via standard evaluation.
+#'   supplied bare expression or from the supplied value via standard
+#'   evaluation.
 #'
 #' @param ... passed on to \code{as_units}
 #'
 #' @param mode if \code{"symbols"} (the default), then unit is constructed from
-#'   the bare expression supplied. Otherwise, if\code{mode = "standard"},
+#'   the expression supplied. Otherwise, if\code{mode = "standard"},
 #'   standard evaluation is used for the supplied value This argument can be set
 #'   via a global option \code{options(units.set_units_mode = "standard")}
 #'
 #' @export
 #' @rdname set_units
+#' @seealso \code{\link{as_units}}
 set_units <- function(x, value, ...,
   mode = getOption("units.set_units_mode", c("symbols", "standard"))) {
   
--- a/R/set_units.R
+++ b/R/set_units.R
@@ -25,8 +25,11 @@ set_units <- function(x, value, ...,
   
   if (missing(value))
     value <- unitless
-  else if (match.arg(mode) == "symbols")
+  else if (match.arg(mode) == "symbols") {
     value <- substitute(value)
+    # if(class(value) == "character")
+    #   stop("Please set `mode = standard` to supply character strings as units")
+  }
   
   if (is.null(value))
     drop_units(x)
--- a/R/set_units.R
+++ b/R/set_units.R
@@ -31,8 +31,6 @@ set_units <- function(x, value, ...,
     if(is.numeric(value) && value != 1)
       stop("The only valid number defining a unit is '1', signifying a unitless unit")
     
-    if(identical(value, quote(unitless)))
-      value <- 1
   }
   
   if (is.null(value))
--- a/R/set_units.R
+++ b/R/set_units.R
@@ -27,8 +27,12 @@ set_units <- function(x, value, ...,
     value <- unitless
   else if (match.arg(mode) == "symbols") {
     value <- substitute(value)
-    # if(class(value) == "character")
-    #   stop("Please set `mode = standard` to supply character strings as units")
+    
+    if(is.numeric(value) && value != 1)
+      stop("The only valid number defining a unit is '1', signifying a unitless unit")
+    
+    if(identical(value, quote(unitless)))
+      value <- 1
   }
   
   if (is.null(value))
--- a/R/set_units.R
+++ b/R/set_units.R
@@ -28,9 +28,8 @@ set_units <- function(x, value, ...,
   else if (match.arg(mode) == "symbols") {
     value <- substitute(value)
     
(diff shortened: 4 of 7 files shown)


t-kalinowski:
Hi @edzer, this isn't quite ready for merge, but it's ready for review. The documentation could use a few tweaks, and a few more tests might be worthwhile.

Sometime next week (if you respond favorably), I'll also submit a PR for adding a `valid_udunits()` and `valid_udunits_prefixes()` functions.

t-kalinowski:
Also, I'm having second thoughts about allowing scalar characters into set_units, specifically strings "hours" "days" .. etc for difftime unit conversion / creation

t-kalinowski:
#60

t-kalinowski:
After taking a closer look, in so far as I can tell, the only functionality this PR breaks without providing a replacement is the interchange between class `difftime` and class `units` objects. 

Conceptually, there is some overlap between the two classes. It is possible to track units of time using the `package:units` infrastructure or the `difftime` methods, with the exception that difftime is limited to time only, where as `units` can have more complicated units where time is just one component of the unit.

With that in mind, I feel like conversion between `units` and `difftime` should be done only explicitly, via something like `as.difftime()` and `as.units()`

`base::as.difftime()` unfortunately is not a generic. What do you think of doing what `dplyr` did with `setdiff()`, defining a generic that masks the base function, setting the default to the base function, and defining the approprite method. something like
```
as.difftime <- function(x,...) UseMethod("as.difftime")
as.difftime.default <- base::as.difftime 
as.difftime.units <- function(x, units = c("auto",  "hours", "days"...) { }
```

edzer:
I am not in favor of masking base functions; units has so far offered `as_difftime` for this purpose. I'm also not happy that we now mask `%*%`, but that one would otherwise simply drop units. Of course, many other functions drop them, like `lm`, so another approach might be desirable, e.g. provide a `%*units%` or so.

t-kalinowski:
ok, I like`as_difftime`. 

I think this is mostly good to merge. The travis build fails because the vignette still needs to be updated, but all the tests otherwise pass (on my computer at least).

edzer:
This is major large overhaul, it'll take me some time to fully comprehend it. I now see a couple of regressions from what we had, in particular:
```r
> library(units)
> make_unit(".001 degree_C*m*m")/make_unit(".001 m")
1 .001 degree_C*m*m/.001 m
```
and
```r
> m
1 m/s
> set_units(1, m)
1 m/s
```
More later.

edzer:
Also, we no longer get symbols, as we had before:
```r
> set_units(1, degree_C)
1 °C
```

t-kalinowski:
Thanks for taking a look. 

In order to keep things simple and predictable, 
`make_units` only accepts bare symbols
`parse_units` only accepts strings

So, the first example part
```
make_unit(".001 degree_C*m*m")
```
doesn't work, because `make_units` in this implementation only accepts symbols, not strings. So instead, we would write (also, please note the plural **s** in 'make_unit**s**':
```
make_units(.001 * degree_C * m * m)
# 0.001 degree_C*m^2
```

Alternatively, to pass a string, we would use parse_units (again, please note, the plural s)
```
parse_units(".001*degree_C*m*m")
# 0.001 degree_C*m^2
```


`set_units` is a little wonky because it's primarily for convenience and interactive use, so it operates both under standard evaluation and then non standard evaluation as a fallback. It first tries to evaluate the supplied expression, and if it resolves to a unit object, it uses that. If not, then it tries non-standard evaluation on the supplied expression and tries to construct a unit object out of that.
So 
```
m <- make_units(m/s)
set_units(1, m)  # since m is an object of class units, it uses that
# 1 m/s
rm(m)
set_units(1, m)
# 1 m  # m does not evaluate to a unit object, so it falls back to inspecting the expression
```
is expected behavior by design. The documentation should perhaps make this more clear.


Regarding the automatic conversion of unit name to unit symbols , e.g. 
degree_C -> °C
I just submitted another commit adding this feature.
now:
```
make_units(degree_C)
# 1 °C
```

t-kalinowski:
It would be fairly straightforward to add an option to `set_units` controlling if SE is done first, or if NSE only is performed. Not sure would it should be called. `NSE_only = FALSE`?

edzer:
I consider that feature creep - both cases should result in a clear error.

t-kalinowski:
Sorry, not quite following. Which of these lines do you think should result in an error?
```
m <- make_units(m/s)
set_units(1, m)  
# 1 m/s
rm(m)
set_units(1, m)
# 1 m  
```

t-kalinowski:
Trying to update the examples about the automatic unit name to unit symbol conversion, I noticed that it doesn't consistently work if the unit name has a prefix. I'm not sure how far down the technology stack this goes, but it behavior that udunits exhibits: 

``` r
library(udunits2)
#> udunits system database read from /Library/Frameworks/R.framework/Versions/3.4/Resources/library/udunits2/share/udunits2.xml
library(magrittr)

ud.get.symbol("grams")
#> [1] "g"
ud.get.symbol("kilograms")
#> [1] "kg"
ud.get.symbol("micrograms")
#> [1] ""

# indeed, prefixes not being resolved to symbols correctly seems to be the norm
x <- units::valid_udunits_prefixes()
#> udunits system database read from /Library/Frameworks/R.framework/Versions/3.4/Resources/library/udunits2/share
paste0(x$name, "grams") %>% sapply(ud.get.symbol)
#> yottagrams zettagrams   exagrams  petagrams  teragrams  gigagrams 
#>         ""         ""         ""         ""         ""         "" 
#>  megagrams  kilograms hectograms  dekagrams  decigrams centigrams 
#>        "t"       "kg"         ""         ""         ""         "" 
#> milligrams micrograms  nanograms  picograms femtograms  attograms 
#>         ""         ""         ""         ""         ""         "" 
#> zeptograms yoctograms 
#>         ""         ""

# and vice versa
paste0(x$symbol, "g") %>% sapply(ud.get.name)
#>           Yg           Zg           Eg           Pg           Tg 
#>           ""           ""           ""           ""           "" 
#>           Gg           Mg           kg           hg          dag 
#>           "" "metric ton"   "kilogram"           ""           "" 
#>           dg           cg           mg           µg           ng 
#>           ""           ""           ""           ""           "" 
#>           pg           fg           ag           zg           yg 
#>           ""           ""           ""           ""           ""
```

I'm not sure this is worth working around in this package. Seems an issue more appropriate for udunits2 or perhaps the underlying C library

edzer:
```r
m <- make_units(m/s)
set_units(1, m)  
# 1 m/s
```
should be an error.

Regarding the symbols: we shouldn't go beyond what udunits does, here; if udunits doesn't give a symbol, we stick to the name.

t-kalinowski:
I agree on the 2nd point about name to symbol conversion. 

Regarding `set_units`, I'm hesitant to convert that to an error. Generally, users won't be working with a custom xml units database, and it seems like quite an annoyance preventing users from using any R variable name that might conflict with a unit name or symbol in the udunits2 xml database. 

I agree that making `set_units` more predictable is a good idea (the current implementation was an attempt to match the current features as much as possible). What do you think about doing away with the NSE completely, and requiring only unit objects (just like `units<-` does currently)?

so, this kind of use would go away completely:
``` set_units(1, m/s) # Error ``` 
instead, users would have to explicitly construct units, eg:
``` set_units(1, make_units(m/s)) ```

**Comment by t-kalinowski** [comment]
set_units in that case would become very simple: 
```
set_units <- function(n, un) {
  units(n) <- un
  n
}
```
 all the error checks are already in `units<-`

**Comment by edzer** [comment]
`set_units` was introduced to make setting units easy, also in pipes, and using NSE. If users want to do this in two steps, they can use `make_unit` and `units<-`. For `set_units`, 

```r
set_units <- function (n, un, allow_user_defined = FALSE) {
    un <- .eval_units(substitute(un), allow_user_defined = allow_user_defined)
    units(n) <- un
    n
}
```
is enough. It makes the habit of
```r
m = make_unit(m)
```
obsolete. It should document that `un` needs to be composed of `udunits` units, unless `allow_user_defined` is FALSE, but will not evaluate in a parent environment.

t-kalinowski:
I like that. 

What do you think of this:
```
set_units <- function (n, ...) {
    units(n) <- make_units(...)
    n
}
```

edzer:
Nice and short. But why are these two equal?
```r
> make_unit(m^2/km)
1 m
> make_unit(m^2/m)
1 m
```
(seems to be also wrong in `units` on CRAN)

t-kalinowski:
I took a quick look, seems to be originating from here:
https://github.com/r-quantities/units/blob/7e9192648c59a72ac236d420ff45f258f1a5fd60/R/arith.R#L72-L73

e1 gets a correct value of 0.001 m, but then e2 also gets to be 0.001 m. 

Thats about as far as I can dig into it for the time being

edzer:
Thanks - good catch!

edzer:
I think that this
```r
> set_units(1:10, 20.11669506 * m)
Units: m
 [1]  1  2  3  4  5  6  7  8  9 10
```
should either give an error, or multiply the values 1:10 with the length of an indian chain in m.

t-kalinowski:
I really like the idea of keeping `set_units` as a simple wrapper around `units<-`. My preference is to change the behavior of either `make_units` or `units<-`

What do you think this should return?
```
m20 <- make_units(20.11669506 * m)
units(m20)$numerator
```

some other scenarios to think consider:
```
z <- y <- x <- 1:3

units(x) <- m20   
x

units(y) <- make_units(m)
units(y) <- m20
y

units(z) <- make_units(m)
units(z) <- units(m20)
z
```

What should `x` `y` and `z` be?

My initial preference: a numeric multipliers should not be accepted in `make_units`. class `.symbolic_units` should only contain symbols. the concept of a unit

t-kalinowski:
Is there a need to create units like `make_unit(20 * m)`? I'm having a hard time conceptualizing the idea of "20 meters" as a unit. Using it in a sentence is confusing: e.g., "The river bank is 5 20 meters long".

What do you think of forcing the returned object from `.eval_units` to have a value of 1, and throwing a warning if not already?

edzer:
I agree; an error would be more appropriate. We have infrastructure in place (which we might need to integrate with this new scheme) for install_conversion_* to define units not available in the udunits database.

t-kalinowski:
What do you think about throwing a warning in `units<-` here, if `as.numeric(value) != 1`
https://github.com/r-quantities/units/blob/7e9192648c59a72ac236d420ff45f258f1a5fd60/R/conversion.R#L38

t-kalinowski:
Hi @edzer, any follow up thoughts on this? It'd be good to get this wrapped up while it's still somewhat fresh in memory.

edzer:
I know and you're right; I'm under a few other deadlines right now, will need about a week.

t-kalinowski:
Sounds good.

edzer:
How, in the current overhaul, do we enter a unit like `hours since 2017-08-20`?

You could argue these are not clean, SI, units, as they have a unit and an offset, but we do need them. For instance, 99% of all climate model data is stored in NetCDF files that have time units described in [this format](http://cfconventions.org/).

edzer:
`make_unit("hours since 2017-08-20")` -- sorry, I was not looking at your latest branch.

edzer:
I now see
```r
> set_units(10, make_unit("m"))
10 1
```

edzer:
I integrated all this in https://github.com/r-quantities/units/tree/overhaul , and corrected tests, docs and vignettes to get a clean check again. Next, I'll check reverse dependencies, to check whether submission to CRAN will break other packages.

edzer:
`set_units` sets a _unit_, it doesn't set _units_ (see [here](https://www.bipm.org/en/publications/si-brochure/chapter1.html)). We now have a mix of *_unit and *_units functions, without clear reason why. I suggest, since we're overhauling, to
* only provide *_unit functions
* rename `units` into `unit` (clashes with `grid::unit`), `units<-` into `unit<-`
* rename the `units` attribute to `unit`
* keep the package name and the class name to `units` (`grid` has a class `unit`).

We also need to think about error checking, since e.g. `symbolic_unit("m s -1")` is just accepted, and doesn't result in anything meaningful, where for instance
```r
> x = symbolic_unit("m s-1")
> 3*x + set_units(4, m/s)
7 (m s-1)
```
Functions should only do what they advertise.

t-kalinowski:
Thanks @edzer! 

I agree, the special time+origin handling of udunits is tricky since its not strictly a pure SI units, and for that reason it breaks certain semantics. We have already good support for time _duration_ units, like seconds, minutes, hours. A "unit" like `"hours since  2017-08-20"` is not so much a unit, as it is a unit with a reference origin. And these duration+reference origin units are only convertible with other duration+origin "units". 

I wonder if a separate set of functions for this special type of unit+origin makes sense. Something like: 
```
time_unit <- function(string, ...) {
  symbolic_unit(string, ...)
}

set_time_unit <- function(n, tm, ...) {
  units(n) <- time_unit(tm, ...)
  n
}
```
Or, perhaps, this use case is different enough that it might make sense to make this a separate class, like `"udunits_time_instant"`.

In any case, this will need to be well documented, and will need to be clear that this is as a special, additional capability of udunits that is exposed, but that since these are not strictly "units" not all things will work. e.g.,
```
set_time_unit(1, "hours since 1990-01-01") + set_units(60, minutes) # ERROR
```

Similarly, arithmetic doesn't always make sense. For example, adding two dates does, conceptually, does not make much sense. Base R reflects that by throwing an error:
``` r
as.Date("1990-01-01") + as.Date("1990-01-02")
#> Error in `+.Date`(as.Date("1990-01-01"), as.Date("1990-01-02")): binary + is not defined for "Date" objects
```

But in `units` this operation is performed, returning an answer that doesn't quite make sense (to me) 
(this is with CRAN units, so won't work on the current branch)
``` r
set_units(1, "days since 1999-10-30") + set_units(1, "days since 1999-10-31")
#> 3 (days since 1999-10-30)
```


I admittedly haven't worked with NetCDF files before. When working with these files, what kind of functionality is desirable? Is it important that arithmetic is performed by `udunits`, or is it ok to just get users into base R classes `Date` or `POSIXct` as quickly as possible?

t-kalinowski:
Re units vs. unit:

To my ear, units sounds the most natural in most contexts. Most sentences that discuss units as an attribute of a number use the form ending in s, units. e.g.,
+  x is in units of meters
+  x is in units of meters/second


The only use of the singluar "unit" that you would use in a sentence is when directly referring to a single defined unit, like "volt" or "second" (not as an attribute of a number).

This reflects the current function family to a certain extent:
`make_units`, `parse_units`, `set_units`, `units`, `units<-`, `as_units`. 
The only exception to this right now is `symbolic_unit`, which, is intentionally different to (hopefully) make clear that it is intended for creating a single unit like "meter", and is not intended for creating compound units like "meters/second"

t-kalinowski:
The whole problem of tracking time is quite thorny, and I'm not sure it should be the purview of units. It means adding support for units of time duration and time instants, and arithmetic between the two where it makes sense, and throwing errors where it doesn't make sense. This functionality is already present in base R with classes `difftime`, `Date` and `POSIXct`, and then also rich functionality in other packages like `lubridate`.

I think the best way to solve this to make clear that `units` is good for doing arithmetic and conversion between well defined quantities, like seconds, hours, minutes, and then encourage users to leverage other resources when they want to do calendar operations, like work with "months" or specific instants in time, and provide easy conversion functions and documentation for that. We already have `as.difftime`, `as.Date` and `as.POSIXct`. Perhaps this should be it's own man page?

edzer:
Allow me to humbly disagree with most your points:
* `m/s` and `m` are two units, but `m/s` is a single unit of measure (see the bipm link: our ears are typically sloppy); the plural does not come from having more than one dimension.
* the `?units` page of package `base` uses `units` in most cases where it should have been `unit`; I must have gotten carried away reading that
* adding a duration to a time point makes perfect sense; try `Sys.time() + difftime(Sys.time() , Sys.time() + 10)`; your example `set_units(1, "days since 1999-10-30") + set_units(1, "days since 1999-10-31")` also makes sense if you think of them as durations, brought to a common offset.
* I suggest to not introduce new functions for time, but rely as much as possible (and hold pkg units as shallow as possible) on udunits, to avoid confusion with the wider community relying on that library. We had this problem in the past, tried to be clever and got the degree_F <--> degree_C conversion wrong.

t-kalinowski:
Thank @edzer. Reading the link you shared, I'm coming around to your point of view. I see that they refer to "m/s" as the unit for speed, using singular unit. However, does that usage of the singular extend to usage of functions, particularly when using them with verbs, e.g. read aloud `set_unit(x, m/s)`:  "Set unit of x to m/s" vs "Set units of x to m/s".  The singular still sounds off to my ear, but I'm guessing that will dissipate with time, and in any case I agree we should be consistent with the usage in the SI definition.

If we do rename all the functions, that would be quite a large change. Is it worth it? One of the nice things about the current implementation is that it extends base R `units`. Although, at this point, that might introduce more confusion than it's worth since `units<-.difftime` accepts only strings in `value`, while `units<-.numeric` and `units<-.units` accepts only `units` objects.

Also, if we do change everything to singular, we probably need a new name for `symbolic_unit`, since it won't be clear anymore what the difference is between `parse_unit` and `symbolic_unit` is anymore.

re: `symbolic_unit("m s -1")`
I don't know the best way to approach this. Compound units like this are not intended for `symbolic_unit`, however, strictly speaking, this is won't lead to any incorrect result. If the symbolic unit is not parseable, a warning is given. In this case, no warning is issued because the string is parable. In this case, errors are thrown correctly when attempting addition or subtraction between incompatible units.  Also, multiplication and division results are still correct, with the caveat that unit simplification isn't performed correctly. 

About the only way I can think of to remedy this to throw a warning if the string passed to `symbolic_unit` contains a space or a `*` or `/` symbol, saying something to the effect of: "Your unit string appears to be a combination of multiple other units, are you sure you shouldn't be using parse_units?

t-kalinowski:
I've been reading the full [PDF](https://www.bipm.org/utils/common/pdf/si_brochure_8_en.pdf) and I think you are right, singular unit is more appropriate.

edzer:
Here is my proposal:
* we keep methods `units` and `units<-`, since these are generics in base which we extend; this is a mistake that we didn't make, and it is in the spirit of R to extend existing methods; in addition, `unit` would collide with `grid::unit` (and is not a generic)
* we rename `verb_units` into `verb_unit` as it is more precise, and removes the currently confusing mix
* `symbolic_unit` errors on anything but a single unit.

I'm proposing to do this during this overhaul, since we're changing API already, and want to avoid doing that twice (other CRAN packages reuse this package).

t-kalinowski:
First bullet: I'm ambivalent. My initial preference is for consistency. If we're renaming everything to `verb_unit` then I feel like `unit` and `unit<-` should be part of the family. Do people regularly attach grid? Is that collision likely in practice?

I agree with the 2nd bullet. 

I agree with the 3rd bullet. How do you propose we implement? Check for one of `\\s\\d*/-` ?

t-kalinowski:
actually, I think I agree with keeping `units` and `units<-`, mostly because it scares me a little thinking about how much existing code it might break for a relatively minor semantic reason, without much gain in functionality.

Related, thinking about `base::units<-`, the semantics are quite different from `units::units<-` in that `value` is expected to be a string in the former. I wonder if, to be consistent, `units::units<-` should also accept a string for `value`, by default passing on the string to `parse_unit`.

t-kalinowski:
After being away from it for a week and reading over all the changes, this whole interface feels too complicated. 

What if we got rid of creating units with bare symbols completely? Then we'd get rid of the whole confusion about NSE and `ud_units` in one go. The primary functions would be `units` and `units<-`, with `set_units` as a pipe friendly version of `units<-`. The primary usage would be:
``` 
units(x) <- "m/s"
# or
set_units(x, "m/s")
```
for special cases or power users, we would export `make_unit` and `symbolic_unit` for creating a unit from a string. Both would work with strings under standard evaluation only.
```
make_unit("m/s")
symbolic_unit("m") / symbolic_unit("s")
```

This type of interface would be simpler, and also more consistent usage of `base::units<-`.

t-kalinowski:
since the primary interface for creating would involve just a string, it would make sense to do some automatic identification of implicit exponents in the string. Something like this:
``` r
are_exponents_implicit <- function(s) {
  s <- trimws(s)
  has <- function(chr, fixed = TRUE) grepl(chr, s, fixed = fixed)
  !has("^") & !has("*") & !has("/") & has("\\s", fixed = FALSE)
}

x <- c( "m*s",  "m/s",  "m s")
are_exponents_implicit(x)
#> [1] FALSE FALSE  TRUE
```

t-kalinowski:
automatic conversion of unit names to symbols and other options to the current `parse_units` might be controllable with a global option, using a default function argument of something like `getOption("units.auto_convert_names_to_symbols", TRUE)`

PR #70, opened 2017-11-06
overhaul unit creation

t-kalinowski:
This is a first pass at some of the discussion points. The documentation and examples haven't been updated, more of a concept at this point.


t-kalinowski:
This PR is orthogonal, mostly, to the unit vs units discussion. 

The thinking here is, there are too many ways to create units between all the `make_units` `parse_units` `set_units` `symbolic_units` functions. This seeks to simplify things by reducing the total number of exported functions.

I used `Date` creation for inspiration with this. The primary way to create `Date` objects is to use the coercion form, `as.Date`. That seems neat, and I think it might work here as well. So the primary unit creation function would be `as_units`. 

The complete user-facing family of functions would be `units` `units<-` (from base), `as_units` `set_units`. 
For backward comparability, `make_unit` is kept as an alias for `as_units.character`.

With this, I'd also consider not exporting `symbolic_unit()`.

Check out the added test for example usage.

edzer:
This looks like a good direction.

I propose to drop the `units` vs `unit` argument, as there is a logic in having `units<-` and `set_units` as a pair (the second being the pipe-friendly version of the first).

But I think `set_units` should keep NSE, at least as an option, for the sake of usability. Also, the idea of having a global option for going to symbols or not is a good one.

t-kalinowski:
Hi @edzer,

I just added a lot of commits to this branch. Still a WIP. 

The two big changes here are 
1) providing character and language methods for the generic `as_units`, and then relying on `as_units` to automatically do the conversion in `units<-` and `set_units()` and 
2) simplify (remove) most of the unit creation function arguments,
   +  `allow_user_defined` is removed in favor of a single new function `define_new_symbolic_unit()` (in line with #72). 
   +  `auto_backtick` is no longer provided as an option. auto backticking seems pretty robust, and for edge cases or power users, we now export `as_units.call` so users can parse the strings themselves and pass the expression if need be.
    +  `auto_convert_name_to_symbol` is now controlled via a global option

The best place to quickly get sense of usage with these changes is the examples here: 
https://github.com/t-kalinowski/units/blob/5dc48e0fe6eb2ee2317cefa1052abb537442483c/R/make_units.R#L46-L148
 
With this direction, we might not need to export `symbolic_unit` anymore.

t-kalinowski:
Ok, ready for review.

t-kalinowski:
The one thing that doesn't work with this new scheme is the substitution of `log` for `ln` `lg`, etc based on the base. I'm not sure the best way to handle this.

t-kalinowski:
I took a first pass at supporting the existing `log` processing in the new `as_units` engine. 

Refactoring the log processing generic method is beyond the scope of what I want to tackle with this PR, but just to start the discussion: I think that the `log` functionality should be refactored to:
*  add support for log10 and log2 (both are generics)
*  `log(., base = exp(1))` calls should be printed as `ln(.)` when printing units, and `log10()` and `log2` should be printed asis. The current `lg()` and `lb()` should be removed (I don't think they are intuitive and add more confusion than they are worth. `log10` and `log2` are clearer).

t-kalinowski:
Hi @edzer, 

Have you had a chance to look over this? 
It passes all build checks now. 

The only part of this that I'm unsure about is the use of a global option in `set_units`, but otherwise I think this is an improvement.

edzer:
Will do; there are a few other duties I have to get out of the way first.

R/set_units.R
```diff
@@ -0,0 +1,52 @@
...
+  
+  units(x) <- as_units(value, ...)
+  x
+}
+
+#' @rdname set_units
+#' @export
+set_units.difftime <- function(x, value, ...) {
+  units(x) <- value
+  x
```
edzer: This does not make sense, as it returns an object of class `difftime`. I think we should return `as_units(x)` rather than x (I will make this change to branch overhaul)

t-kalinowski: My thinking here is that people will think of `set_untis` as a simple, pipe friendly alias for `units<-`. This assumption is fine to make for almost all the uses, except for the case of converting units on difftime objects.

t-kalinowski: I think that coercion from difftime to units should be done explicitly by users, with `as_units`... `set_units` is more for assigning or converting units

edzer: No, then you keep living in two units systems; `set_units` should move to the world of the units package. Also, the argument value needs to be character, the names incompatible with udunits and so on. Different world.

t-kalinowski: ok. In that case, we perhaps should no longer make it a generic. I think it should error out for difftime objects, and users should be encouraged to coerce explicitly with as_units. having set_units to do automatic conversion between difftime to units feels too magical.

edzer: I agree to remove it.

Issue #91, opened 2018-01-20
Error: In ‘g/kg’ the numeric multiplier ‘0.001’ is invalid


samuel-rosa:
In version 0.4-6 of `units`, when setting the measurement unit of a quantity as `g/kg`, the quantity used to be multiplied by `0.001` to achieve simplification, returning a unitless quantity. Now, with version 0.5-1, the following error message is thrown:

```r
> library(units)
> x <- 1.3
> set_units(x, g/kg)
Error: In ‘g/kg’ the numeric multiplier ‘0.001’ is invalid. 
Use `install_conversion_constant()` to define a new unit that is a multiple of another unit.
> 
> install_conversion_constant(from = "g", to = "kg", const = 1/1000)
> install_conversion_constant(from = "kg", to = "g", const = 1000)
> 
> set_units(x, g/kg)
Error: In ‘g/kg’ the numeric multiplier ‘0.001’ is invalid. 
Use `install_conversion_constant()` to define a new unit that is a multiple of another unit.
```

Although the behaviour in the current version seems a bug that may need a fix (I am not sure about this), I would like to know how to avoid the simplification of units. The unit `g/kg` is extensively used in my field (soil science) and it is very important that it be kept as is when used with a quantity.

I have seen that `units` throws the error message above every time we use a unit composed of physically related units: `g/kg`, `mL/L`, and so on.

I think that there is some discussion about this in http://www.sciencedirect.com/science/article/pii/S0263224117303305.

t-kalinowski:
In the example above, are `g` and `kg` referring to different things? For example, grams arsenic / kg soil? Or are they referring to the same thing (g soil / g soil). 

If it's the former, i wonder if instead of `install_conversion_constant`, this problem would be better addressed through `install_symbolic_unit()`, or something like it.

edzer:
Whatever it is, this is a regression introduced in 0.5-0. 

SI units primarily don't address the _of what_ question. The enhancement @samuel-rosa points to is a proposal to further specify dimensionless units, so that e.g. `rad`, counts of persons, and `g/g` units are no longer convertible.

t-kalinowski:
Agreed, I was just thinking of where and how this usage might be accommodated. I don't think simply removing the error check in `as_units` is the right solution.

samuel-rosa:
[...]

Issue #89, opened 2018-01-18
deprecate install_conversion_constant


edzer:
It is clear from #71, #84, and my experiments in the udunits branch that user-defined units are better handled by the udunits2 C library than by R. I removed the R layer of installing user-defined units; this removes the flexibility to define _any_ R function for conversion, which has the advantage that we don't have to verify that functions are meaningful (monotonous), which we didn't. Also, the udunits2 C api does give, I think, all meaningful operations, which we can now directly interface.

I propose to drop `install_conversion_constant` in favor of `install_conversion_scale`, which complements `install_conversion_offset`. The problem with `_constant` is that it doesn't make clear whether the constant refers to a scaling or an addition of a constant; `_scale` and `_offset` are used in the C API. 

I also propose the new semantics:
```r
install_conversion_scale("m", "thousandm", 1000)
```
creates `thousandm` where X `thousandm` equal 1000 X `m`. This is exactly opposite to what `install_conversion_constant` now does. Also, `thousandm` should not already be bound to a unit, and `m` should be an existing unit.

Similarly:
```r
install_conversion_offset("K", "my_temp_C", 273.15)
```
creates the new `my_temp_C` as values in `K` + 273.15.

The function profile is `(from,to,value)`, and returns to=from*value or to=from+value.

Enchufa2:
Are you planning to drop the udunits2 dependency in the next release?

edzer:
If you and @t-kalinowski both agree this is a good idea, yes.

Enchufa2:
Completely agree. I would like to redesign the C/C++ part, but I may not have time depending on your roadmap for the next release. Anyway, as it is, it works, so the redesign can wait.

edzer:
Agreed; something possibly more urgent now: I managed to get memory leaks to zero, but still see, when units (udunits branch) is loaded:
```
edzer@gin-pebesma:~/git$ R --no-restore --no-save -q -d valgrind
==24394== Memcheck, a memory error detector
==24394== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==24394== Using Valgrind-3.12.0 and LibVEX; rerun with -h for copyright info
==24394== Command: /usr/lib/R/bin/exec/R --no-restore --no-save -q
==24394== 
> library(units)

Attaching package: ‘units’

==24394== Conditional jump or move depends on uninitialised value(s)
==24394==    at 0x40300A9: ???
==24394==    by 0x142FE99F: ???
[...]

PR #94, opened 2018-01-23
alternative solution to #89; #91

t-kalinowski:
With this implementation, the symbolic units are preserved so they can be printed correctly, and assigning a unit like "mg/kg" no longer changes the numeric value of the target vector.

Also, fixes the regression mentioned here (and elsewhere):
https://github.com/r-quantities/units/pull/64#issuecomment-339036324

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -5,6 +5,7 @@
 NULL
 
 .onLoad = function(libname, pkgname) {
+  # can this be lozy loaded on first access attempt instead?
 	udunits_init(file.path(.get_ud_xml_dir(), "udunits2.xml"))
 }
 
--- a/R/init.R
+++ b/R/init.R
@@ -5,7 +5,7 @@
 NULL
 
 .onLoad = function(libname, pkgname) {
-	udunits_init(.get_ud_xml_dir())
+	udunits_init(file.path(.get_ud_xml_dir(), "udunits2.xml"))
 }
 
 .onAttach <- function(libname, pkgname) {
--- a/R/make_units.R
+++ b/R/make_units.R
@@ -285,12 +285,29 @@ units_eval_env$lb <- function(x) base::log(x, base = 2)
 
 units_eval_env$`/` <- function(x, y) {
   # don't cancel units, just combine them
+  if (!inherits(x, "units") || !inherits(y, "units"))
+    return(x/y)
+  
   units <- structure(list(
-    numerator = c(units(x)$numerator, units(y)$denominator), 
-    denominator = c(units(y)$numerator, units(x)$denominator)), 
+    numerator =   c(units(x)$numerator,   units(y)$denominator), 
+    denominator = c(units(x)$denominator, units(y)$numerator)), 
     class = "symbolic_units")
-  val <- as.numeric(x) / as.numeric(y)  #if(is.finite(y <- as.numeric(y))) y else 1
+  val <- drop_units(x) / drop_units(y)  
+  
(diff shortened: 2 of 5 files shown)


t-kalinowski:
Not quite complete yet. With units like `"20 * g /kg"`, the "20" gets discarded, which instead it should throw an error. It should be a simple 1 line fix, but there is no way to write any unit tests for this until #95 gets resolved.

t-kalinowski:
ready for review

edzer:
Thanks! I'm reluctant to merge this, as we still see conversion:
```r
> set_units(1, kg/g) + set_units(2)
1.002 kg/g
> set_units(1, kg/g) + set_units(1, "rad")
1.001 kg/g
```
and simplification when created from expressions involving units:
```r
> set_units(1, kg) / set_units(2, g)
500 1
```
I think simplification is wanted, and the way to go is that of the Flater paper. In case of a unitless unit, we need to keep where it comes from: `m/m` can be added to `rad`, but `s/s` not. I guess this calls for an additional field in the `symbolic_units` class.

There are a couple of other things in this PR that I will integrate manually.

As usual, I'll be happy to discuss.

edzer:
So it seems I now have merged this; however, I didn't plan to. What a day.

Enchufa2:
:D I apologise for my share. I feel I'm abusing a little bit your units day.

edzer:
No, no, everything is fine - I see where you're going, very useful.

I reverted back to before the merge of this PR, merge master again and did a `git push -f` in branch `udunits`. This means that this PR has not been merged. Sorry for the mess!

t-kalinowski:
haha. 

For what it's worth, here is a quick response to your concerns:
This is designed to address the specific case of creating a unit like "ug/kg", so that you could do, for example, conversion to "mg/kg", without simplification to `unitless` occurring (exactly the use case in the linked to issue, and also a use case that I think is very common.) 

The other use cases I think are mostly orthogonal to this PR.

I think that it's entirely reasonable that simplification occurs when multiplying or dividing per-existing unit objects outside the `as_units` constructor (and, more often than not, simplification is desirable, I think). To prevent simplification during arithmetic, a solution is to encourage users to create custom units like "ug_Arsenic / kg_soil". The issue of arithmetic between conceptually incompatible `unitless` units, like rad and kg/g, I think is also solved by recommending the creation of custom units.

t-kalinowski:
Also, the current solution that allows for construction of units like "mg/kg", changes the value of the number vector on `units(x) <- "ug/kg"`, and I don't think that's right. This should always return `TRUE`
``` R
identical(x, drop_units(set_units(x, un)))
# should return true for all `x` and `un`
```

edzer:
We now have
```r
> a = 1:10
> units(a) <- "ug/kg"
Error: In ‘ug/kg’ the numeric multiplier ‘1e-09’ is invalid. 
Use `install_conversion_constant()` to define a new unit that is a multiple of another unit.
> units(a) <- "ppm"
> a
Units: ppm
 [1]  1  2  3  4  5  6  7  8  9 10
> a + set_units(1)
Units: ppm
 [1] 1000001 1000002 1000003 1000004 1000005 1000006 1000007 1000008 1000009
[10] 1000010
```
I'm not so optimistic about the ability to add new units into udunits C API: they are either unitless and can then be added to any other unitless, or compatible to one of the existing units. A "new" unit is only meant for something really new (not compatible with the 7 SI base units); if you have weights of Arsenic, it is a weight, and can be added to any other weight.

Issue #95, opened 2018-01-23
passing pointer from R -> Cpp -> R not working


t-kalinowski:
``` r
library(units)
#> udunits system database from /usr/local/share/udunits
#> 
#> Attaching package: 'units'
#> The following object is masked from 'package:base':
#> 
#>     %*%
units:::ud_are_convertible("20 * s", "seconds since 1970-01-01")
#> Error in R_ut_are_convertible(R_ut_parse(as.character(u1)), R_ut_parse(as.character(u2))): external pointer is not valid
```

Enchufa2:
Let me take a look,

edzer:
Remove the spaces around the `*`:
```r
> units:::R_ut_parse("20*s")
<pointer: 0x5571c247b230>
```

edzer:
... or write no, one or multiple spaces to imply multiplication:
```r
> units:::R_ut_parse("20*s")
<pointer: 0x55d67debea80>
> units:::R_ut_parse("20s")
<pointer: 0x55d67d8cb650>
> units:::R_ut_parse("20 s")
<pointer: 0x55d67c83a820>
> units:::R_ut_parse("20  s")
<pointer: 0x55d67c84e290>
```
all mean the same.

Enchufa2:
Yeap, but the thing is... why are we returning `XPtr`s at all? Each time you get a pointer to a `ut_unit`, it is a clone of an internal object. At R level, we should work with names and symbols only.

Enchufa2:
Check #98, please.

edzer:
Of course we can do without, as that is what we've done so far. It's an experiment, and still in an early phase. With Xptr, you can do
```r
> units:::R_ut_format(units:::R_ut_parse("20*s"))
[1] "20 s"
```
where you wouldn't have to redo the parse for every print. Also, convert after checking conversion is possible, can now be done without parsing both units _twice_.

Now before you are going to tell me that these operation have negligible cost, please provide some pointers to large computations where indeed doing units did not have a cost overhead. [This](http://r-spatial.org/r/2017/11/13/perp-performance.html) post gave me worries that units may involve overhead,  but I still have to go to the bottom of it.

t-kalinowski:
re @edzer's first suggestions for alternative syntax: all good suggestions but if `"20 * m"` is not valid then `ut_is_parseable()` should reflect that.

Enchufa2:
[...]

PR #98, opened 2018-01-23
remove XPtr

Enchufa2:
Udunits works by maintaining a map of objects in a system of units. Each time we add a new unit, a new object is created and a *clone* returned. Each time we parse an existing unit, a *clone* to an existing object is returned. Two implications: 1) there is no need to save pointers, nor return them to R; 2) we should call *_free for every pointer returned by a udunits function. In other words, every function exported to R should 1) do its thing, 2) clean up.

Therefore, this is what this PR contains:

- No need to parse from R: R_ut_parse removed.
- inst/include/units.h removed.
- All functions returning an XPtr now return nothing.
- All functions working with pointers clean them up.
- Unused functions working with XPtrs (R_ut_root and the like) commented out.


Enchufa2:
One _problem_ with `R_ut_are_convertible`: we have to think how we want to use it. Namely, is a call to `R_ut_is_parseable` always required before calling it? If the answer is "yes", we should set the error message handler to ignore any errors here. If not, we should handle the errors inside (e.g., by calling `R_ut_is_parseable` inside `R_ut_are_convertible`.

In other words, with this PR, as it is, we have

```r
units:::ud_are_convertible("20 * s", "seconds since 1970-01-01")
#> ut_are_convertible(): NULL unit argument[1] FALSE
```

Because `"20 * s"` is not properly formatted and the parser fails returning a NULL. So do we want an error here, or just `FALSE` because a call to `ud_is_parseable` should be required in the first place?

t-kalinowski:
```
> ud_is_parseable("20 * s")
[1] TRUE
```

Enchufa2:
@t-kalinowski Are you using this PR? I see:

```r
units:::ud_is_parseable("20 * s")
#> [1] FALSE
```

t-kalinowski:
I haven't tried out this PR where you removed the pointers.... I likely won't get a chance to really play with it until the weekend.

I don't yet have an opinion about removing pointers... as Edzer mentioned, it's an experiment in early stages. Having pointers allows you to more easily do some pretty interesting things as far as building mini DSL's. I thought originally we might want that for installing user defined units, but we might not need them after all.

Also, I think `units:::ud_is_parseable("20 * s")` __should__ return `TRUE`, and it's ud_are_convertable that should be more accommodating. `ut_trim`?

Enchufa2:
> Also, I think units:::ud_is_parseable("20 * s") should return TRUE, and it's ud_are_convertable that should be more accommodating. ut_trim?

The fact is that udunits cannot parse "20 * s"  even after trimming. The only way would be to trim out that spaces *before* passing the string to udunits.

Enchufa2:
Regarding this PR:

As I said in #95, I'm concerned about performance too, but there's no gain in the current implementation. Moreover, I'd say there's a penalty: the overhead of allocating `XPtr` here and there without actually using them.

We may improve performance only if we cache unit objects and are able of identifying and retrieving them fast, without having to parse strings again. In R, this wouldn't be efficient; it should be done at C++ level.

So my vote is to accept this PR to simplify things for now. In a future redesign, we could encapsulate the units system into a C++ object and expose a clean interface without having to call udunits functions directly.

edzer:
I started this as a branch from `master`, to experiment in, and now it mixes good things with experimental things. If you want to progress with master, write a PR that pulls in the good things from this branch into master, rather than revert back the things I am experimenting with, leaving me with additional burden to carry on with the experiment.

Enchufa2:
Ok, fair enough. Closing...

Enchufa2:
If you don't mind, I would like to keep this branch for my own experimentation.

edzer:
Sure!

PR #109, opened 2018-02-20
use NextMethod for concatenation

Enchufa2:
As in #107, this change doesn't affect `units` behaviour, but enables integration into `quantities`.


Enchufa2:
Sorry, it seems it's not working as I expected (for `quantities`). The `errors` method is not dispatched for some reason. Let me fix it.

PR #110, opened 2018-02-21
use NextMethod for concatenation (2)

Enchufa2:
The previous patch apparently worked, at least for `units` alone, but effectively `NextMethod` wasn't dispatching any method after `c.units`. I still don't understand why, and it may be a bug in R, because the same code has a different behaviour depending on the scope (i.e., whether the methods are defined in the global environment or exported from packages; see [this reprex](https://github.com/Enchufa2/dispatchS3dots) for more details).

However, I managed to circumvent this issue with this patch. It's based on calling `c` once more if any argument was modified (units changed), and finally calling `NextMethod` cleanly. It is not an ideal solution, because this means that, if any unit changes, `c.units` is called _twice_. But AFAIK, it may be the only solution. And if there is a bug in R, then this is the only backwards-compatible solution.

### DIFF
--- a/R/misc.R
+++ b/R/misc.R
@@ -1,21 +1,28 @@
 #' @export
 c.units <- function(..., recursive = FALSE) {
-  modified <- FALSE
   args <- list(...)
-  u = units(args[[1]])
-  for (i in seq_along(args)[-1]) {
-    if (!inherits(args[[i]], "units"))
-      stop(paste("argument", i, "is not of class units"))
-    if (units(args[[i]]) != u)
-      modified <- TRUE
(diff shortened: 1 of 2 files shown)


Enchufa2:
[Tomas Kalibera thinks](https://stat.ethz.ch/pipermail/r-devel/2018-February/075620.html) that there is no way to replace (unnamed) arguments in dots for `NextMethod`. So apparently this workaround is the only way.

Enchufa2:
I've added this commit to this PR because the `Summary.units` method needs the same mechanism, so I have isolated it in a helper function called `.convert_to_first_arg`.

PR #116, opened 2018-02-27
fix #115: sign drops units

### LINKED ISSUE #115: sign of units
Should the `sign` function preserve units? It does now, but I think it shouldn't.

Enchufa2:
Simple patch that returns ASAP. 

Under this _principle_, this whole `Math.units` can be improved in terms of performance (the same applies to `Ops.units`). I've been tempted, but I love, as you said once, _fancy things_, such as [this](https://github.com/r-quantities/errors/blob/master/R/math.R#L26), and you may not like it. So I leave the _greater good_ for another day, another discussion. :-)

### DIFF
--- a/R/math.R
+++ b/R/math.R
@@ -20,6 +20,9 @@ Math.units = function(x, ...) {
   if (.Generic == "sqrt")
     return(x^0.5)
   
+  if (.Generic == "sign")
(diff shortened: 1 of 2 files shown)


edzer:
Thanks for saving the fanciness for another day!

PR #121, opened 2018-02-28
fix multiplication and division (2)

Enchufa2:
I've just realised that @mailund's PR broke `NextMethod` dispatch by calling `as.numeric(NextMethod())`. In `quantities`, units methods are always called in the first place, because if the operation needs a unit conversion, errors for that conversion must be propagated. Thus, enclosing `NextMethod` in `as.numeric` throws the errors out.

In the same way, we cannot rely on conversions using the C API (i.e., calling `convert`) directly, because errors are not propagated either. All conversions must ~take place at R level~ be wrapped inside a registered method (i.e., `units<-`), to be able to intercept the call and retrieve the conversion applied from `quantities`. See changes in `.simplify_units` for further details.

### DIFF
--- a/R/symbolic_units.R
+++ b/R/symbolic_units.R
@@ -125,6 +125,7 @@ as.character.symbolic_units <- function(x, ...,
   # before removing matching units.
 
   drop_ones = function(u) u[ u != "1" ]
+  class(value) <- "units"
   
   new_numerator <- drop_ones(sym_units$numerator)
   new_denominator <- drop_ones(sym_units$denominator)
@@ -136,7 +137,8 @@ as.character.symbolic_units <- function(x, ...,
       str2 <- new_denominator[j]
 
       if (are_convertible(str1, str2)) {
-        value <- convert(value, str1, str2)
+        attr(value, "units") <- str1
+        units(value) <- str2
         delete_num <- c(delete_num, i)
         new_denominator <- new_denominator[-j]
         break
@@ -147,5 +149,5 @@ as.character.symbolic_units <- function(x, ...,
   if (length(delete_num) > 0)
     new_numerator <- new_numerator[-delete_num]
   
-  as_units(value, .symbolic_units(new_numerator, new_denominator))
+  as_units(drop_units(value), .symbolic_units(new_numerator, new_denominator))
 }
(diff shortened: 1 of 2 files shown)


R/symbolic_units.R
```diff
@@ -136,7 +137,8 @@ as.character.symbolic_units <- function(x, ...,
       str2 <- new_denominator[j]
 
       if (are_convertible(str1, str2)) {
-        value <- convert(value, str1, str2)
+        attr(value, "units") <- str1
+        units(value) <- str2
```
edzer: I don't see the point of this, since `units<-.units` calls `convert()`; I find this even more obfuscating than it was.

Enchufa2: But the point is that `units<-.units` can be intercepted by `units<-.quantities`, which is able to check whether there was a conversion, and applies it to the errors (see [here](https://github.com/r-quantities/quantities/blob/master/R/units.R#L15)).

Enchufa2: So I may have been unclear: conversions still take place through `convert` eventually, but this call must be enclosed in something that can be intercepted and managed from the outside.

edzer: I trust you on this one!

mailund:
I was completely unaware of the interplay between `quantities` and `units`. That might also have implications for the Issue I just wrote (#123). I think we can handle all conversion in R in the way I describe there, but I don't know which consequences it will have when combining different ways of "extended" numerics. I only had `units` and scalars in mind.

Enchufa2:
No problem, this is pretty new, and I'm starting to identify the cornerstones that must be preserved to guarantee that all the packages play nicely together. So far, these are the rules that I more or less confidently can establish:

- Rely on `NextMethod` whenever possible, having in mind that the input may contain other classes (`unclass` and the like must be avoided), and the output may contain other attributes (at this point, classes can be dropped, but attributes must be preserved).
- Conversions can be as complex as required, but must be wrapped around `units<-` and called through this interface. In this way, `quantities` is able to place itself in the first place in the dispatch stack, and retrieve and propagate the conversion.

R/symbolic_units.R
Enchufa2: Thanks!

PR #122, opened 2018-02-28
fix plot and hist (2)

Enchufa2:
Sorry for this, I was **completely wrong** about this. I said that the call was not being modified, so it was not necessary to provide the labs, but **of course it is** being modified! The user types `plot(something)` and the method **adds** the labels... I should have checked it more carefully.


edzer:
Well, I should have reviewed more carefully!

Issue #75, opened 2017-11-19
units API overhaul


edzer:
See #70 

Still TODO:
```
help of ?set_units: ... ; mention NSE


# This should not occur:
> units(b) = "10 m"
> b
Units: ^10*m
[1] 1 2 3


# This should be an error:
> b = 1:3
> units(b) = "10*m"
Warning message:
In ‘10 * m’ the numeric multiplier ‘10’ was discarded. 
The returned unit object was coerced to a value of 1.
Use `install_conversion_constant()` to define a new unit that is a multiple of another unit. 


# This should not occur:
> set_units(1:3, as_units("m"))
Units: 1
[1] 1 2 3
```

t-kalinowski:
This is a tough edge case. Here, the call `as_units("m")` is passed on to as_units.call. `all.vars` returns a zero length vector, since there are no names or symbols in there, just a function and a constant (string). 

The best way to guard against this is explicitly I think. if `mode == "symbols`, add an check like this
```
 if (mode == "symbols") {
  value_expr <- substitute(value)
  value <- try(force, silent = TRUE)
  if (length(all.vars(value_expr) == 0L) && 
      (is.character(value) ||  inherits(value, "units") || inherits(value, "symbolic_units)
   stop("set_units requires a bare expression. set mode = "standard" to pass a bare string or units object")
  value <- value_expr  
}

```

or, better yet, move the error into as_units.call (where `x` is the expression)

```
constants_supplied <- vapply(x, is.character, logical(1L))
if(any(constants_supplied))
  stop("please supply only R symbols, not bare strings")
```

t-kalinowski:
if we add that check into `as_units.call`, we'd have to modify `as_units.expression` to unwrap the list before passing on to the call, like this: 
```
as_units.expression <- function(x, ...) as_units.call(x[[1]], ...)
```

t-kalinowski:
I added some more commits to related to this to #64, and trimmed a few other branches in the meantime. I think you'll have to merge then into the r-quantities/units/overhaul branch though, as I don't think I can push to that branch (please correct me if I'm mistaken)

t-kalinowski:
```
> # This should not occur:
> units(b) = "10 m"
> b
> Units: ^10*m
> [1] 1 2 3

**Comment by t-kalinowski** [comment]
I addressed most of the TODO's, except this one:
```
# This should not occur:
> units(b) = "10 m"
> b
Units: ^10*m
[1] 1 2 3
```
[...]

PR #149, opened 2018-06-17
convert implicit_exponents -> explicit_exponents

t-kalinowski:
Working towards #75

### DIFF
--- a/R/make_units.R
+++ b/R/make_units.R
@@ -228,7 +228,7 @@ as_units.character <- function(x,
     implicit_exponents <- are_exponents_implicit(x)
   
   if(implicit_exponents)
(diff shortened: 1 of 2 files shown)


Enchufa2:
We should add tests to our PRs if some behaviour is changed. Note that, if the checks pass both with and without the PR, it's difficult to figure out what exactly the PR addresses, especially when dealing with regexps. ;-)

t-kalinowski:
Agreed. 

This PR is just a refactor to consolidate the code paths, no change in input or output expected.  

The code coverage is decreasing mostly because many lines have been deleted... the few new added lines are covered.

edzer:
I never liked that code - thanks!

PR #154, opened 2018-06-21
implement as.list

Enchufa2:
It doesn't help with aggregations, as we discussed, but it's beneficial to fix things like the following:

```r
lapply(set_units(1, m), function(i) i)
[[1]]
[1] 1
```

It should be instead

```r
lapply(set_units(1, m), function(i) i)
[[1]]
[1] 1 m
```

This is because `lapply` and others call `as.list` internally, which drops units by default. Also in this PR:

- A small fix in `as.data.frame` to set proper column names and behave as the default method in that sense.
- Tests for `as.data.frame` and `as.list`.


edzer:
Looks good!

PR #160, opened 2018-07-06
close #159: clarify drop_units documentation and describe alternatives

### LINKED ISSUE #159: drop_units() stopped working for values which are not of class "units"
drop_units() stopped working for values which are not of class "units" as of version 0.6-0 from CRAN.

Up to version 0.5-1, drop_units() used to work with values of any type (returning the value x unchanged if class(x) did not contain "units"). In version 0.6-0, drop_units() has been converted into an S3 method without a default implementation (#118). For non-"units" arguments it therefore results in an error message, e.g.:

``` r
units::drop_units(1)
#> Error in UseMethod("drop_units"): no applicable method for 'drop_units' applied to an object of class "c('double', 'numeric')"
```

Created on 2018-07-04 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

The previous behaviour could be restored by providing a drop_units.default() implementation, e.g.:

``` r
drop_units.default <- identity
```

Enchufa2:


edzer:
Thanks!

PR #161, opened 2018-07-10
use specified udunits header locations

yasirs:
AC_CHECK_HEADER was not using the specified --with-udunits2-include argument or the UDUNITS2_INCLUDE environment variable.

The CPPFLAGS were never updated, and the UD_CPPFLAGS never used in AC_CHECK_HEADER. This fixes the situation.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -26,9 +26,11 @@ AC_ARG_WITH([udunits2-include],
 		[udunits2_include_path=$withval])
 if test [ -n "$udunits2_include_path" ] ; then
    UD_CPPFLAGS="-I${udunits2_include_path}"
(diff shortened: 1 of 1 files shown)


edzer:
Thanks a lot!

Enchufa2:
There is duplicated code now. [This line](https://github.com/yasirs/units/blob/a8dba68a4a6849cd492dd2d858f9e2174404c826/configure.ac#L54) already did that. We should revert this PR.

Enchufa2:
@yasirs Have you checked the line I linked in my previous comment?

yasirs:
@Enchufa2 Yes I agree. I had only looked at the CRAN version.

Enchufa2:
Thanks, reverting then.

PR #167, opened 2018-08-13
Guard suggests usage in examples, vignettes and tests

Enchufa2:
Suggested packages were not properly guarded in examples, vignettes and tests, and `R CMD check` was failing if those suggests were not installed.

This PR includes commits from `convert_to_base`, which needs to be merged in `master`.


Enchufa2:
Please also note that your local repo was still in `convert_to_base` in the last 2 commits.

Issue #176, opened 2018-09-26
support for decibels?


t-kalinowski:
I'm not entirely sure how this should work. 

Short of a holistic solution, a first step might be sensible printing if you recreate the unit manually (so the `lg(1)` below might instead be `dB`
``` r
library(units)
#> udunits system database from /usr/share/udunits
signal_power <- set_units(10, W)
noise_power <- set_units(2, W)

snr <- 10 * log(signal_power / noise_power, base = 10)
snr
#> 6.9897 [lg(1)]
```

(as a side note, it would also be good to add support for `log10` and `log2` at some point, so the above `log(..., base = 10)` wouldn't be necessary.

<sup>Created on 2018-09-26 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

edzer:
`udunits` apparently knows about the `dB` [prefix](https://en.wikipedia.org/wiki/Decibel):
```r
> set_units(1:3, dBm)
Units: [dBm]
[1] 1 2 3
> set_units(1:3, dBV)
Units: [dBV]
[1] 1 2 3
```
I guess your example should result in `[dBW]`?

Enchufa2:
It is not a prefix, but a unit:

```r
units:::R_ut_format(units:::R_ut_parse("dBm"))
#> [1] "0.1 lg(re 0.001 W)"
```

means 0.1 log referenced to 1 mW.

t-kalinowski:
Yeah, dBm is close, but not entirely applicable for dB. dB is really a ratio on a log scale. dBm is a special case of dB with specific known value in the denominator (1 milliwatt). I don't think it would make sense to convert snr (signal-to-noise ratio) to dBm in this case.

Enchufa2:
One way to get dB:

```r
library(units)
#> udunits system database from /usr/share/udunits
install_conversion_constant("lg(re 1)", "dB", 1)
signal_power <- set_units(10, W)
noise_power <- set_units(2, W)
set_units(signal_power / noise_power, dB)
#> 0.69897 [dB]
```

Enchufa2:
@edzer: Perhaps `units` should define that conversion automatically? And we should fix the `log` operation, because `lg(1)` means nothing for udunits:

```r
units:::R_ut_format(units:::R_ut_parse("lg(1)"))
#> Error in units:::R_ut_parse("lg(1)") : 
#>  Error in function R_ut_parse: string unit representation contains unknown word
```

[...]

PR #177, opened 2018-09-27
Support for logarithms

Enchufa2:
Initial implementation to address issues in #176.


Enchufa2:
@edzer, @t-kalinowski: I think this PR is ready. Please, take a look. Summary:

- Logarithms now work via udunits2.
- `BW` (bel-Watts) and others are defined by udunits2, but not `B` (bel). With this PR, by default `B` is added, unless a user-provided XML database defines some `B` already or it's disabled via the `define_bel` option.
- Some tests and docs (see `help(Math.units)`) about the behaviour of the logarithm have been added.

If you agree with the changes so far, we can merge this PR and close #176. Maybe the best is to avoid any automatic conversion (to dBW, etc.) at the moment: we can always revisit the discussion in #176 if the users demand that feature.

t-kalinowski:
By the way, I find your recent comments in #176 convincing, and I think it's probably OK to do automatic conversion.

Enchufa2:
Nooo... Did we convince each other? :D :D :D

Enchufa2:
Ok, I've added a couple of tests for the decibel too. May I squash and merge this, @edzer?

edzer:
Yes, very good - go ahead. Great work!!

Issue #187, opened 2019-01-21
add drop_units.data.frame?


kendonB:
This might make dealing with models a little smoother. Ref: https://www.r-spatial.org/r/2018/08/31/quantities-final.html#fitting-linear-models-with-quantities

kendonB:
To be clear, I would expect `drop_units.data.frame` to remove the units attribute from all columns of a `data.frame`.

PR #191, opened 2019-02-17
adding data.frame method for drop_units

kendonB:
addresses https://github.com/r-quantities/units/issues/187


R/make_units.R
```diff
@@ -440,3 +440,12 @@ drop_units.units <- function(x) {
   x
 }
 
+#' @export
+drop_units.data.frame <- function(.data){
```
Enchufa2: An S3 method should be consistent with its generic: in this case, the argument must be `x` instead of `.data`.

Enchufa2:
Also, not required, but it would be nice if you could provide a small test for this new functionality and/or update the NEWS.

Enchufa2:
@edzer Please, have a look and merge this if you are ok with the addition.

Enchufa2:
@edzer Kind reminder. ;-)

edzer:
Thanks for reminding!

PR #196, opened 2019-03-09
Allow headers outside standard path

### LINKED ISSUE #195: Installation fails on MacPorts
When I tried to install this package under MacPorts, the configure script found udunits2/udunits2.h, but the compilation failed. A closer look revealed that the compiler was searching for /usr/include/udunits2/udunits2.h, whereas MacPorts stores the actual header in /opt/local/include/udunits2/udunits2.h.

In configure.ac, the `AC_CHECK_HEADER` macro searches for headers in the default locations used by the compiler. If udunits2/udunits2.h is found, configure.ac assumes that the headers are stored under /usr/include. This will cause problems on any system with header files stored outside /usr/include (such as MacPorts).

mjwoods:
Resolves #195 .

We can avoid changing the search path of the compiler, and instead use conditional preprocessor statements to include either `<udunits2.h>` or `<udunits2/udunits2.h>`, based on whichever is found by `configure`.


Enchufa2:
Thanks for the PR. The installation fails on Windows now. Please, see the
AppVeyor output.

mjwoods:
I'm not sure how to see the AppVeyor output, so I'm guessing what the problem is on Windows. Hopefully it is fixed now.

edzer:
See [here](https://ci.appveyor.com/project/edzer/units), after downloading aftifacts you'll see in `00install.out`: 
```
* installing *source* package 'units' ...

   **********************************************
   WARNING: this package has a configure script
         It probably needs manual configuration
   **********************************************


** libs

*** arch - i386
rm -Rf units.dll RcppExports.o io.o udunits.o
"c:/R/bin/i386/Rscript.exe" "../tools/winlibs.R"
mkdir -p ../inst
rm -Rf ../inst/share
cp -r ../windows/udunits-2.2.20/share ../inst/
C:/Rtools/mingw_32/bin/g++  -I"c:/R/include" -DNDEBUG -DUDUNITS2_DIR=0 -I../windows/udunits-2.2.20/include -I"c:/RLibrary/Rcpp/include"        -O2 -Wall  -mtune=generic -c RcppExports.cpp -o RcppExports.o
RcppExports.cpp:4:35: fatal error: ../inst/include/units.h: No such file or directory
 #include "../inst/include/units.h"
                                   ^
compilation terminated.
make: *** [c:/R/etc/i386/Makeconf:215: RcppExports.o] Error 1
ERROR: compilation failed for package 'units'
* removing 'C:/projects/units/units.Rcheck/units'
In R CMD INSTALL
```

Enchufa2:
The problem is that you renamed `inst/include/units.h` as `inst/include/units.h.in`, and made it a config file, but the configure script is not executed on Windows, so `inst/include/units.h` is never created. Please, use `Makevars` to define `UDUNITS2_DIR` instead.

Enchufa2:
Perfect, thanks! Merging now.

PR #197, opened 2019-03-10
Force units' finalizers before freeing the system

Enchufa2:
[This](https://github.com/r-quantities/units/compare/unload?expand=1#diff-2711d8988b4edf6562a0668d7e873c13R27) trick is not very elegant, but it's actually needed. The thing is, each time a unit is wrapped into an Rcpp XPtr, a weak finalizer is registered (which calls to `ut_free` eventually). This is a problem, because we can't control when they are executed. As a result, the package cannot be unloaded cleanly, because there's [a segfault](https://travis-ci.org/r-quantities/units/jobs/504166374#L1376) (builds succeeded because we hadn't [warnings_are_errors](https://github.com/r-quantities/units/compare/unload?expand=1#diff-354f30a63fb0907d4ad57269548329e3R9) set). The segfault happened because 1) `ut_free_system` was called upon unloading, and then 2) `ut_free` was called randomly in a weak finalizer. This call to `gc()` forces the execution of those pending finalizers before freeing the unit system.

And you'd be wondering why we didn't see this until now. The answer is that there was a typo [here](https://github.com/r-quantities/units/commit/956062cf13641bc20d205d4b4066437d558582b8#diff-2711d8988b4edf6562a0668d7e873c13R25), so `udunits_exit` was not being called.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ r:
   - devel
   #- oldrel
 
+warnings_are_errors: true
+
 addons:
   apt:
     packages:
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -26,11 +26,11 @@ R_convert_doubles <- function(from, to, val) {
 }
 
 R_ut_new_dimensionless_unit <- function(name) {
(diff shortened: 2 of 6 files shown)


edzer:
Looks good - thank!

PR #209, opened 2019-05-25
Fix #208: validate units before installation

### LINKED ISSUE #208: Validate unit names before installation
See #180. Any attempt to install a new unit name/symbol with

- leading or trailing numbers;
- arithmetic operators;
- white spaces;
- ... am I missing something?;

should issue an informative error.

Enchufa2:
Current checks:

- Leading and trailing numbers.
- Arithmetic operators.
- Intermediate white spaces.

Am I missing something?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * fix support for weights with units in `weighted.mean`; #205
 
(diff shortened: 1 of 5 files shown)


edzer:
Looks good & makes a lot of sense - thanks!

PR #230, opened 2020-03-16
Remove use of isFALSE to allow use of R < 3.5

### LINKED ISSUE #229: Use of isFALSE requires R>=3.5
In the recent release 0.6-6, the isFALSE function is used in `R/symbolic_units.R` which causes calls to units to fail on R < 3.5 since isFALSE was introduced in 3.5. Is your intent to no longer support older version of R or is this an error?

Either way, I am happy to create a pull request bumping the minimum support R version or removing the call to isFALSE. I am using units for the [yggdrasil](https://github.com/cropsinsilico/yggdrasil) package and noticed the error on CI jobs during a release [today](https://travis-ci.org/github/cropsinsilico/yggdrasil/builds/663192839).

langmm:
Closes #229


Enchufa2:
I think it's more safe and clear if we just port the function and keep the syntax. Also, I've bumped the version and updated the NEWS. Many thanks!

PR #233, opened 2020-03-26
close #232: fix replacement of parts of units

### LINKED ISSUE #232: unit conversion with rbind (on data.frames)
Hi, 
I just came across the following issue, in which the automatic unit conversion does not apply. 
The reprex aims to `rbind` two data.frames which share the same column 'dt' of class 'units'. The unit however differs and the first one is used during `rbind`. Most probably this is rather a "rbind"-related issue but I thought it is worth sharing it here...

``` r
library(units)
#> udunits system database from /usr/local/share/udunits

# dummy data
h <- data.frame(dt = set_units(24, "h"))
min <- data.frame(dt = set_units(1440, "min"))

# binding both data.frames works, but silently gives wrong unit conversion results
# issue: first unit dominates the other 

# ... here: hour
rbind(h, min)
#>         dt
#> 1   24 [h]
#> 2 1440 [h]
# ... here: mins
rbind(min, h)
#>           dt
#> 1 1440 [min]
#> 2   24 [min]
```

<sup>Created on 2020-03-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Enchufa2:
With this patch,

```r
x <- set_units(24, h)
x[2] <- set_units(1440, min)
x
#> Units: [h]
#> [1] 24 24

x[2] <- set_units(3, m)
#> Error: cannot convert m into h

h <- data.frame(dt = set_units(24, "h"))
min <- data.frame(dt = set_units(1440, "min"))
rbind(h, min)
#>       dt
#> 1 24 [h]
#> 2 24 [h]
rbind(min, h)
#>           dt
#> 1 1440 [min]
#> 2 1440 [min]
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -39,6 +39,6 @@ SystemRequirements: udunits-2
 License: GPL-2
 URL: https://github.com/r-quantities/units/
 BugReports: https://github.com/r-quantities/units/issues/
-RoxygenNote: 7.0.2
+RoxygenNote: 7.1.0
 Roxygen: list(old_usage = TRUE)
 Encoding: UTF-8
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -2,7 +2,9 @@
 
 S3method("[",mixed_units)
 S3method("[",units)
+S3method("[<-",units)
 S3method("[[",units)
+S3method("[[<-",units)
 S3method("units<-",logical)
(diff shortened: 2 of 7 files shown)


edzer:
Great work!! I looked at the rbind.data.frame code for a long time, but couldn't bring it back to the missing `[<-` and `[[<-` problem!

PR #247, opened 2020-06-10
Add vctrs methods

lionel-:
Part of r-lib/vctrs#1130


lionel-:
Currently there is rudimentary support via fallbacks. I think #248 first needs to be addressed to get robust tidyverse support for mixed units (with or without vctrs methods).

Enchufa2:
Merged, thanks!

Issue #244, opened 2020-05-11
Allow restore of database


hughjonesd:
Slightly related to #243, I'd like to restore the original database sometimes (e.g. on unload of a package that uses `install_` functions, or when switching between different unit "namespaces"  - e.g. in my case between LaTeX and CSS ideas about lengths). At present I can remove what I've installed but cannot easily install what I have remoced.

Enchufa2:
Yeap, we could add a function to re-read and re-populate the database.

Enchufa2:
Maybe even a mechanism to load external user-provided databases.

PR #254, opened 2020-11-07
Add load_units_xml()

### LINKED ISSUE #243: Allow local versions of database
If I want to use - e.g. - "pt" for points, not pints, I can do that with `install...` functions. But now suppose a package does that, but another package wants to use "pt" for pints. I don't think at present that will work. It would be nice if different namespaces could have their own versions of the udunits database.

Enchufa2:
New function to load a user-provided XML file or to reload the default one if no path is provided. Closes #243 (not supported) and #244 (addressed here).


Enchufa2:
@edzer Kind reminder.

edzer:
Great work - thanks!!

Issue #252, opened 2020-10-26
Suggestion for new function: keep_units()


vorpalvorpal:
There are quite a few functions out there that don't cope well with units objects. I find myself not infrequently writing:
`
x <- set_units(2, mg);
set_units(some_function(drop_units(x)), mg)
`
or, even worse, when I don't know what the units will be:
`
x_units <- units(x);
set_units(some_function(drop_units(x)), x_units, mode = "standard")
`

I wonder if a convenience function along the lines of the following would be useful:

```r
keep_units <- function(FUN, args, arg_unit = NULL){
  fix_units <- function(arg){
    arg_unit_i <- units(arg)
    if(is.null(arg_unit)){
      arg_unit <<- arg_unit_i
    }
    if(!identical(arg_unit, arg_unit_i)){
      arg <- set_units(arg, arg_unit, mode = "standard")
    }
    arg <- drop_units(arg)
    return(arg)
  }
  args <- rapply(args, fix_units, classes = "units", how = "replace")
  do.call(FUN, args) %>%
    set_units(arg_unit, mode = "standard") %>%
    return()
}
```

which would allow `keep_units(some_function, list(x))` which is quite a bit shorter and, I think, easier to read.

Enchufa2:
Could you provide a couple of examples of such functions, please?

Enchufa2:
Your approach seems overly complicated. Something like the following would be sufficient:

```r
keep_units <- function(FUN, x, ..., unit=units(x)) {
  set_units(do.call(FUN, list(x, ...)), unit, mode="standard")
}
```

unless I'm missing something, and that's why asked for some examples of functions that drop units. And anyway, it is always better to provide methods for such functions whenever possible.

vorpalvorpal:
The most recent one I came across was smwrBase::fillMissing, but I have had the same problem with other functions. 

The idea of my function was that some functions might take multiple different unit objects as arguments. Thus:
```
a <- set_units(1, g)
b <- set_units(1, mg)
# the following works correctly
a + b
# so this should work as well
keep_units(FUN, list(a, b))
```

Enchufa2:
Addition works correctly because there are rules encoded for that in the corresponding method. But how could an arbitrary function possibly work with multiple units? What should `keep_units` do with them?

vorpalvorpal:
What I find really useful about the units package is that I can take data
from different sources and work with them without having to convert them
into common units. So I might receive some data in mg/l and other data in
[...]

PR #255, opened 2020-11-07
New helper keep_units

Enchufa2:
Simple helpers to address use cases described in #252 more easily.


edzer:
I can see the point of `keep_units()`, but find `adapt_units` a bit namespace clobbering, as it doubles `set_units` essentially wrapping its `mode` argument. Can't we make `set_units` more clever and have it choose `mode` automatically? The problem is really here:
```
> a = set_units(1, m)
> set_units(a, set_units(1, km))
Error: could not find function "set_units"
Did you try to supply a value in a context where a bare expression was expected?
> set_units(set_units(1, m), set_units(1, km), mode = "standard")
0.001 [km]
```
the error is not helpful, and always when I use `set_units` in code, I use standard mode, and I have to look up its documentation because I forgot.

Enchufa2:
Ok, I'll rebase this PR and work on making `set_units` more clever when we resolve the others.

Enchufa2:
I've been thinking about making `set_units` more clever, and it may break current functionality. Suppose we want this to work:

```r
x <- set_units(1000, m)
set_units(x, make_units(km))
```

The problem is that we would need to actually evaluate the `value` argument to check whether it is a `units` or a `symbolic_units` object, to switch to standard evaluation in such a case. But then NSE evaluation is broken depending on the user's environment. E.g.,

```r
km <- set_units(1, l)
set_units(x, km)
```

would fail, because we cannot convert meters to liters.

To try to avoid such breakage, we could 1) try symbolic evaluation, 2) try standard one if the symbolic fails, 3) throw the first error if the second fails too. But then the method gets very inefficient (more `try` expressions in the execution path) and still there are many cases in which both methods would succeed.

So I think that `set_units` should stay the same, even if we drop `adapt_units` from this PR. IMHO, `set_units` never should have allowed NSE for units in the first place. But now we have to deal with it. :)

Issue #248, opened 2020-06-11
Mixed units can't be empty


lionel-:
I see with the dev version:

```r
m = c(set_units(1, m), set_units(2, g), allow_mixed = TRUE)
m[0]
#> Mixed units: Error in names(dn) <- dnn : attempt to set an attribute on NULL
```

lionel-:
They also have trouble with missing indices which are important for tidyverse support:

```r
# Logical NA should recycle to full length
m[NA]
#> Mixed units: Error in UseMethod("units") :
#>   no applicable method for 'units' applied to an object of class "NULL"

# Integer NA should represent one missing element
m[c(NA, 1)]
#> Mixed units: Error in UseMethod("units") :
#>   no applicable method for 'units' applied to an object of class "NULL"
```

This would require a value representing a missing unit vector (in list classes we generally use `NULL` for this).

PR #256, opened 2020-11-07
Fix printing of empty mixed_units

Enchufa2:
Subsetting worked ok. It was the `print` and `str` methods which were buggy for the cases described in #248.


Enchufa2:
Also, I found that the `c.mixed_units` method was defined but not exported. Fixed here too.

Enchufa2:
@edzer Kind reminder.

Issue #211, opened 2019-10-08
set_units(1, unit) and with(ud_units, unit) are not equivalent


moodymudskipper:
The README says :

>The units package comes with a list of over 3000 predefined units,

```
length(ud_units)
## [1] 3248
```

Yet `ud_units$year` doesn't exist and `set_units(1, year)` works.

Could you clarify the relationship between these 2 syntaxes ?

Moreover, how can I access the list of all the spellings supported by `set_units` ? I dug into the code but I bumped into C code.

Enchufa2:
`year` is an alias for `yr`:

```
ud_units$yr
## 1 [yr]
```

We use UDUNITS-2 as a backend, that's why you hit C code. If you want to dug into the complete list of units and aliases supported, you can check the XML files under `system.file("share/udunits", package="units")`.

moodymudskipper:
Great thanks! I'm leaving open because in case you might want to clarify the README

moodymudskipper:
Some things cannot be explained by those lists alone it seems, for example `set_units` accept `kiloPa` and `kPascal` but these seem to be written nowhere, is this logic in the C code ?

Enchufa2:
Prefixes are in `udunits2-prefixes.xml`; units, in all the other files. By the way, you can also inspect the units using `units::valid_udunits()` if you have `xml2` installed.

There's the documentation of the UDUNITS2 library, which is referenced several times, so I don't think that replicating it here is a good idea. Commenting on every single unit UDUNITS2 supports does not scale up either, less so in a README.

Nevertheless, the package has two vignettes, one of which is an R Journal paper. If you think that they can be improved in any of these aspects, feedback is very welcome.

Enchufa2:
The next version will no longer have the pre-computed `ud_units` list, which was soft-deprecated since v0.5-0. The old examples with `with(ud_units, unit)` have been updated to `make_units(unit)`.

PR #259, opened 2020-11-21
remove deprecations, bump minor version

Enchufa2:
This removes `as.units`, `as_cf`, `make_unit` and `parse_unit`, which have been deprecated for a long time (since v0.5, 2018-01-07), and consequently bumps version to 0.7-0. I removed all calls and mentions to them in docs, demos and tests.

I think it's time to remove `ud_units` too, which is unneeded and soft-deprecated since v0.5 as well (and it only generates confusion, e.g. #211), but I wanted to ask you @edzer first.


edzer:
Good idea. What was again the reason to deprecate `ud_units`, do we no longer use it? I mention it in the vignette, that would need updating too.

Enchufa2:
Prior to v0.5-0, `ud_units` was internally used by `set_units` and `as_units`. NEWS for v0.5-0 said:

> `ud_units` is no longer necessary and is soft-deprecated, and may be removed in a future release.

edzer:
OK, let's pull the plug. I'll update that vignette.

Enchufa2:
Ok, plug pulled in the commit referenced above. I've removed:

- the private `sysdata.rda`;
- all the code associated;
- the few tests associated;
- all the examples in docs, demos and the other vignette.

I've converted every appearance of `with(ud_units, someunit)` to `make_units(someunit)`.

edzer:
Great!

PR #273, opened 2021-03-16
Use default pillar_shaft() method for numbers

krlmlr:
This gives underlines for every other block of three digits, and all the other details of default pillar formatting.

Working on making formatting customizable, see https://pillar.r-lib.org/dev/articles/numbers.html for a primer. When merged, units will also benefit.

CC @lionel-.

Includes part of #274.


R/tidyverse.R
```diff
@@ -6,12 +6,11 @@ type_sum.units <- function(x, ...) {
...
 
 type_sum.mixed_units <- function(x, ...) {
-  "mixed_units" 
+  "mixed_units"
 }
 
 pillar_shaft.units <- function(x, ...) {
-  out <- format(unclass(x), ...)
-  pillar::new_pillar_shaft_simple(out, align = "right", min_width = 8)
+  pillar::pillar_shaft(unclass(x))
```
lionel-, line 13: Should dots be passed here?

krlmlr: Good catch!

Enchufa2:
LGTM. Thanks!

PR #294, opened 2021-12-27
integrate ggplot2 scales, closes #164

### LINKED ISSUE #164: Cannot Use ggplot2 with units
I just tried to plot data using ggplot2 with a units vector, and it yielded an error.  Diving into the ggplot2 code, the issue comes down to `range(NULL, units_value)` returning a numeric vector which cannot be compared with a `units` class object.  Specifically, this occurs in https://github.com/tidyverse/ggplot2/blob/8778b48b37d8b7e41c0f4f213031fb47810e70aa/R/range.r#L21-L25 which involves a call to `scales::train_continuous` which has either `NULL` or a numeric vector as the first argument to `range`.

I don't see a way to fix this without either converting the value I'm plotting to numeric (easy, but losing information and losing an opportunity to make a common pair of libraries more usable together) or converting the `units` library to an S4 class, but maybe you see a simpler way to do this?

``` r
library(units)                         
#> udunits system database from C:/Users/Bill Denney/Documents/R/win-library/3.4/units/share/udunits
library(ggplot2)                       
                                       
d <- data.frame(x=set_units(1:3, m),   
y=1:3)                                 
ggplot(d, aes(x=x, y=y)) + geom_point()
#> Don't know how to automatically pick scale for object of type units. Defaulting to continuous.
#> Error in Ops.units(x, range[1]): both operands of the expression should be "units" objects
```

Enchufa2:
@edzer I've adapted the code from `ggforce` and added Thomas as a contributor in the DESCRIPTION. I've called the functions `scale_*_units` instead of `scale_*_unit` to follow the naming conventions. I've also dropped `ggforce` from the DESCRIPTION, demos and vignettes, but preserving the mention in the ACKs (i.e., [this](https://github.com/r-quantities/units/compare/feature/ggplot2?expand=1#diff-94f753ca5edf0983b3a5de83f7d997aeb8bfc70fd68569a25ff9fb01438251d7R450-R451)). Is that ok?


edzer (approved):
Looks good to me, great work!!

Enchufa2:
Thanks, @edzer.

And @krlmlr thanks for your reference implementation in `pillar`. Here's ours, any comments welcome. The only drawback is that both require pillar/units to be **attached**, not only loaded, for the scales to work. Can you think of any way to make ggplot2 find these scales when the packages are only loaded?

krlmlr:
Thanks, good catch:

``` r
data <- tibble::tibble(
  x = pillar::num((1:10) / 100, fixed_exponent = -3, notation = "eng"),
  y = pillar::num((1:10) / 100, scale = 100, label = "%"),
  z = pillar::num(10^(-5:4), notation = "si")
)
data
#> # A tibble: 10 × 3
#>         x     y     z
#>     <eng>     %  <si>
#>  1  10e-3     1   10µ
#>  2  20e-3     2  100µ
#>  3  30e-3     3    1m
#>  4  40e-3     4   10m
#>  5  50e-3     5  100m
#>  6  60e-3     6    1 
#>  7  70e-3     7   10 
#>  8  80e-3     8  100 
#>  9  90e-3     9    1k
#> 10 100e-3    10   10k

ggplot2::ggplot(data, ggplot2::aes(x, y)) +
  ggplot2::geom_point()
#> Don't know how to automatically pick scale for object of type pillar_num/pillar_vctr/vctrs_vctr/double. Defaulting to continuous.
#> Don't know how to automatically pick scale for object of type pillar_num/pillar_vctr/vctrs_vctr/double. Defaulting to continuous.
```



<sup>Created on 2021-12-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I'll try to fix in pillar.

krlmlr:
This is a current limitation, about to file a ggplot2 issue.

For now, should `scale_type()` raise a warning if the package is not attached?

Enchufa2:
> This is a current limitation, about to file a ggplot2 issue.

Great, thanks.

> For now, should `scale_type()` raise a warning if the package is not attached?

That's a very good idea.

Enchufa2:
9ec86ee will do for now. In our case, I used `stop` because the plot would throw a more confusing error anyway. In your case, a warning would be more appropriate. And the good thing is that `scale_type` is not called when the scales are explicitly provided, so the error/warning doesn't trigger. Example:

``` r
library(ggplot2)
mtcars$consumption <- units::set_units(mtcars$mpg, mi / gallon)
mtcars$power <- units::set_units(mtcars$hp, hp)

ggplot(mtcars) +
  geom_point(aes(power, consumption))
#> Error in scale_type.units(x): Variable of class 'units' found, but 'units' package is not attached.
#>   Please, attach it using 'library(units)' to properly show scales with units.

ggplot(mtcars) +
  geom_point(aes(power, consumption)) +
  units::scale_x_units(unit = "W") +
  units::scale_y_units(unit = "km/l")
```



``` r
library(units)
#> udunits database from /usr/share/udunits/udunits2.xml
ggplot(mtcars) +
  geom_point(aes(power, consumption))
```



<sup>Created on 2021-12-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

PR #302, opened 2022-02-11
Windows: update to udunits-2.2.28

### LINKED ISSUE #301: install_unit() that is equivalent to an existing unit - different behaviour on Windows
I would like to be able to install a unit that is equivalent (identical) to an already installed unit. This works fine on macOS and Linux, but fails on Windows

Using the latest version of `{units}` - I've confirmed this with on macOS with the binary CRAN version and the latest github version (built with the latest `{udunits2}` from homebrew), but it fails on Windows.

### macOS

``` r
library(units)
#> udunits database from /usr/local/share/udunits/udunits2.xml

install_unit("foobar", "1 kg", "A foobar")
x <- as_units(5, "foobar")
x
#> 5 [foobar]
units(x) <- "g"
x
#> 5000 [g]
```

### Windows

```r
units::install_unit("foobar", "1 kg", "A foobar")
#> Error in ud_map_symbols(symbol, ut_unit) : Unit already maps to "kg"
#> Backtrace:
#> x
#> 1. \-units::install_unit("foobar", "1 kg", "A foobar")
#> 2. \-units:::ud_map_symbols(symbol, ut_unit)
```

I notice that `{units}` on Windows is built against an old version of `udunits2` ([v2.2.20](https://github.com/rwinlib/udunits)), whereas it is built against the current ([v2.2.28](https://github.com/R-macos/recipes/blob/a4115f28a84432c167ff4c3320f25cffe05cca1c/recipes/udunits)) version on macOS, so I'm guessing the difference might lie there?

jeroen:
Fixes #301

### DIFF
--- a/src/Makevars.win
+++ b/src/Makevars.win
@@ -1,5 +1,6 @@
-PKG_CPPFLAGS = -DUDUNITS2_DIR=0 -I../windows/udunits-2.2.20/include
-PKG_LIBS = -L../windows/udunits-2.2.20/lib${R_ARCH} \
+RWINLIB = ../windows/udunits-2.2.28
+PKG_CPPFLAGS = -DUDUNITS2_DIR=0 -I${RWINLIB}/include
+PKG_LIBS = -L${RWINLIB}/lib${R_ARCH}${CRT} \
 	-ludunits2 -lexpat
 
 all: clean winlibs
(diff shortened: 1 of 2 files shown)


Enchufa2:
Thanks!

PR #335, opened 2022-12-17
Always allow division of identical units

### LINKED ISSUE #310: Unusual units with log(x) that cannot be divided
This may be related to #249 or #292.

I'm working with fitting data on the log scale that simplifies to the below evaluation.  There are three underlying issues here that may be better treated separately.  (If you'd prefer them opened as separate issues, I'm happy to do so.)

Issue 1 is that it seems odd that units would create a unit that it cannot recognize.  I recognize that more is probably happening here, but if it were possible to simplify the units with log transforms as indicated by setting the units to "ln(g)" in the example below, it would work.

Issue 2 is that it can add the units but it cannot divide them.  It would seem that even if units could not be recognized, if they are identical, you could end up with a unitless value, regardless.

Issue 3 is that when dividing log units by themselves, they do not cancel to be unitless.

``` r
x1 <- log(units::set_units(2, "g"))
x1 + x1
#> 1.386294 [ln(re 1 g)]
x1/x1
#> Warning: Could not parse expression: '`ln`(`re` * ^(1) * `g`)'. Returning as a
#> single symbolic unit()
#> Error: '`ln`(`re` * ^(1) * `g`)' is not a unit recognized by udunits or a user-defined unit
x2 <- x1
attr(x2, "units")$numerator <- "ln(g)"
x2 + x2
#> 1.386294 [ln(g)]
x2/x2
#> 1 [ln(g)/ln(g)]
```

billdenney:
Fix #310

This always allows division of units, if they are equal, and it returns a unitless object.

### DIFF
--- a/R/arith.R
+++ b/R/arith.R
@@ -71,18 +71,12 @@ Ops.units <- function(e1, e2) {
     identical(units(e1)$numerator, units(e2)$denominator) &&
     identical(units(e1)$denominator, units(e2)$numerator)
 
-  if (div && identical_units) {
-    # Special cases for identical units which may not be otherwise divisible by
-    # udunits (see #310)
-    if (.Generic == "%/%") {
-      return(set_units(drop_units(e1) %/% drop_units(e2), 1))
-    } else if (.Generic == "/") {
-      return(set_units(drop_units(e1)/drop_units(e2), 1))
-    } else {
-      stop("Unknown division .Generic, please report a bug") # nocov
-    }
-  } else if (mul && inverse_units) {
-    return(set_units(drop_units(e1) * drop_units(e2), 1))
+  if ((div && identical_units) | (mul && inverse_units)) {
+    # Special cases for identical unit division and inverse unit multiplication
+    # which may not be otherwise divisible by udunits (see #310)
+    e1 <- drop_units(e1)
+    e2 <- drop_units(e2)
+    return(set_units(NextMethod(), 1))
   } else if (mod) {
     div <- e1 / e2
     int <- round(div)
--- a/R/arith.R
+++ b/R/arith.R
@@ -46,7 +46,7 @@ Ops.units <- function(e1, e2) {
     return(NextMethod())
 
   eq  <- .Generic %in% c("+", "-", "==", "!=", "<", ">", "<=", ">=") # requiring identical units
-  div <- .Generic %in% c("/", "%/%")
+  div <- .Generic %in% c("/", "%/%")                                 # division-type
   prd <- .Generic %in% c("*", "/", "%/%", "%%")                      # product-type
   pw  <- .Generic %in% c( "**", "^")                                 # power-type
   mod <- .Generic %in% c("%/%", "%%")                                # modulo-type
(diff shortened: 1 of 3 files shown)


R/arith.R
```diff
@@ -60,7 +61,21 @@ Ops.units <- function(e1, e2) {
...
-  if (mod) {
+  identical_units <-
+    inherits(e1, "units") & inherits(e2, "units") &&
+    identical(units(e1), units(e2))
+
+  if (div & identical_units) {
+    # Special cases for identical units which may not be otherwise divisible by
+    # udunits (see #310)
+    if (.Generic == "%/%") {
+      return(set_units(as.numeric(e1) %/% as.numeric(e2), ""))
```
Enchufa2, line 72: We cannot use `as.numeric` in general, because it breaks compatibility with `quantities` (see the check errors). Please use `drop_units` instead in both arguments instead. Then, we can just call `NextMethod()`, and there is no need for branches.

Regarding "unitless", please use a `1` as we do down below in the "prd" branch. Just a matter of consistency.

Enchufa2 (changes requested):
Thanks for the PR! Please see the comments. Additionally, what about the product? If `x1` is a log-unit, shouldn't this work too?

```r
x1 * (1/x1)
```

R/arith.R
```diff
@@ -60,7 +61,21 @@ Ops.units <- function(e1, e2) {
     units(e2) <- units(e1) # convert before we can compare; errors if unconvertible
   }
 
-  if (mod) {
+  identical_units <-
+    inherits(e1, "units") & inherits(e2, "units") &&
```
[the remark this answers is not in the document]
billdenney: No good reason.  Fixed.

R/arith.R
```diff
@@ -60,7 +61,21 @@ Ops.units <- function(e1, e2) {
...
+    identical(units(e1), units(e2))
+
+  if (div & identical_units) {
```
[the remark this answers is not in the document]
billdenney: Fixed

R/arith.R
billdenney: For going to `NextMethod()`, wouldn't it be preferable to have the output as a unitless number (the output I have) rather than a bare number?  That was what I was going for, but maybe you prefer a bare number (or maybe I'm missing something).

billdenney:
The product when the units are inverted should work, too.  Let me think about that a bit...

R/arith.R
Enchufa2: `set_units(NextMethod(), 1)`

billdenney:
I just added a commit so that inverse unit multiplication always works.

R/arith.R
billdenney: Neat!  I just learned how NextMethod works!  Thanks!

R/arith.R
```diff
@@ -60,7 +62,22 @@ Ops.units <- function(e1, e2) {
...
+    inherits(e1, "units") && inherits(e2, "units") &&
+    identical(units(e1)$numerator, units(e2)$denominator) &&
+    identical(units(e1)$denominator, units(e2)$numerator)
+
+  if ((div && identical_units) | (mul && inverse_units)) {
+    # Special cases for identical unit division and inverse unit multiplication
+    # which may not be otherwise divisible by udunits (see #310)
+    e1 <- drop_units(e1)
+    e2 <- drop_units(e2)
+    return(set_units(NextMethod(), 1))
```
Enchufa2, line 79: Nice! Much simpler. :)

`||`, please. ;-)

R/arith.R
```diff
@@ -60,7 +62,22 @@ Ops.units <- function(e1, e2) {
...
 
-  if (mod) {
+  identical_units <-
+    inherits(e1, "units") && inherits(e2, "units") &&
+    identical(units(e1), units(e2))
+
+  inverse_units <-
+    inherits(e1, "units") && inherits(e2, "units") &&
+    identical(units(e1)$numerator, units(e2)$denominator) &&
+    identical(units(e1)$denominator, units(e2)$numerator)
```
Enchufa2, line 72: If you don't mind, let's take the opportunity for a small improvement, which is to call `inherits(e1, "units")` and `inherits(e2, "units")` just _once_ in this function and then reuse the result, because there are too many of these calls already throughout the function.

Enchufa2 (changes requested):

R/arith.R
billdenney: Fixed

R/arith.R
billdenney: I updated this.  It required also making one `if` into an `else if`.

Enchufa2 (approved):
Thanks!

PR #336, opened 2022-12-17
Always allow comparisons to logical or numeric NA

### LINKED ISSUE #308: Feature Request: Allow binary comparisons and arithmetic operators with NA that does not have units
All binary comparisons and arithmetic operators (e.g. `<`, `>`, `+`, `*`, etc.) when performed with an `NA` value result in `NA`.  It would be helpful if the Ops with units allowed NA values to have these comparisons by default.

``` r
units::set_units(1, "m") < NA
#> Error in Ops.units(units::set_units(1, "m"), NA): both operands of the expression should be "units" objects
```

billdenney:
Fix #308

This always allows comparison to numeric or logical NA (and it adds a test for non-NA comparisons, too).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+# development version
+
+* NA to unit comparisons (for logical and numeric NA) will always return NA;
+  #308 @billdenney
+
 # version 0.8-1
 
 * fix `%/%` and `%%` if arguments have different units; #313
--- a/R/arith.R
+++ b/R/arith.R
@@ -56,7 +56,7 @@ Ops.units <- function(e1, e2) {
 
   if (eq) {
     if (!(inherits(e1, "units") && inherits(e2, "units")))
-      stop("both operands of the expression should be \"units\" objects") # nocov
+      stop("both operands of the expression should be \"units\" objects")
     units(e2) <- units(e1) # convert before we can compare; errors if unconvertible
   }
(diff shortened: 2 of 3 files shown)


Enchufa2:
About this one, I'm not convinced either that we should treat `NA` differently. I think too that missing values should have units to allow the comparison, as with any other numeric.

billdenney:
My main use case is that the way I get data is not always clean.  (I'm a heavy contributor to the janitor package for that reason 😄 .)

I will often get data where there is an NA indicator when a measurement is not taken.  That NA will not have units associated.  And, I may want to compare the NA to the median of the distribution.  In my code, I can remove the units to check for NA, but to me, the general idea of "comparison to missing is always missing" holds.

For the use case in the PKNCA library, I currently do many checks for NA.  When trying to add units support, I'd need to do some special handling to allow the library to work with and without units.  Allowing NA comparison would prevent that need.

That said, if it's philosophically not aligned with the package, I can make the more extensive code changes in my other code.

Enchufa2:
Yes, I'm sorry, but philosophically, a data frame column, for example, may be longitude or mass or whatever, and it should have the proper units for comparison independently of whether there is no missing measurement, or some missing measurements, or all missing. Let's keep it in that way for consistency. Thanks for your understanding!

PR #337, opened 2022-12-17
Update to newer CI versions

billdenney:
The PRs I just made failed.  It appeared to me that the failures were due to CI issues and not the PRs themselves.  This updates the CI files to the newest versions from r-lib.  I think that I got the required customizations to install udunits correct.


Enchufa2:
Thanks for the intent, but current checks are fine. `quantities`, `units` and `errors` share YAML definitions, and the second two trigger specific tests for `quantities` too. I like them the way they are. :)

billdenney:
That makes sense.  I saw that you were doing some other interesting stuff in there, and after seeing more, I realized that I didn't pull it all through.

You may want to still consider converting to `actions/checkout@v3` as a more focused change.

Enchufa2:
Thanks, I'll take a look. Anyway the errors we had were due to some exceptional connectivity issue at some stage of `setup-r`, which is in the newest version. In my experience, official actions rarely fail, even the old ones.

PR #338, opened 2022-12-17
Preserve names with unit conversion

### LINKED ISSUE #305: Names are lost when units cancel but not when units do not cancel
In the example below, `names()` are lost when the units cancel, but they are retained when units do not cancel.  I think that the preferred behavior would keep the `names()`.

``` r
library(units)
#> udunits database from C:/Users/Bill Denney/Documents/R/win-library/4.1/units/share/udunits/udunits2.xml
a <- setNames(set_units(1, "kg"), "A")
# names not shown on normal scalar print
a
#> 1 [kg]
# but they are still there
names(a)
#> [1] "A"

b <- setNames(set_units(c(1, 2), "kg"), c("A", "B"))
# names shown on vector print
b
#> Units: [kg]
#> A B 
#> 1 2
names(b)
#> [1] "A" "B"

d <- a/set_units(1, "mol")
d
#> 1 [kg/mol]
names(d)
#> [1] "A"

e <- b/set_units(1, "mol")
e
#> Units: [kg/mol]
#> A B 
#> 1 2
names(e)
#> [1] "A" "B"

# names are lost when units cancel to become unitless (for scalars)
f <- a/set_units(1, "mg")
f
#> 1e+06 [1]
names(f)
#> NULL

# names are lost when units cancel to become unitless (for vectors)
e <- b/set_units(1, "mg")
e
#> Units: [1]
#> [1] 1e+06 2e+06
names(e)
#> NULL
```

<sup>Created on 2022-02-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

billdenney:
Fix #305

Now, names are preserved with unit conversion.  The underlying issue is that the C-connection function `ud_convert` drops the names.  I don't know how to fix that, and it seems to be used only at the location that I found in the code.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# development version
+
+* Names are preserved when doing unit conversions; #305 @billdenney
+
 # version 0.8-1
 
 * fix `%/%` and `%%` if arguments have different units; #313
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -67,7 +67,7 @@
 #' units(a) <- make_units(km/h)
 #' a
 #' # convert to a mixed_units object:
-#' units(a) = c("m/s", "km/h", "km/h")
+#' units(a) <- c("m/s", "km/h", "km/h")
 #' a
 `units<-.units` <- function(x, value) {
   if(is.null(value))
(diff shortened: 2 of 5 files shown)


Enchufa2:
Good catch! Thanks for looking into this. Then let's try to just preserve those names in `ud_convert()`.  Please, try the commit above.

billdenney:
Since it passes the added test, it works as desired (and better than my fix since it works at the source).

PR #353, opened 2023-08-09
remove tolerance from comparisons with logical operators

### LINKED ISSUE #351: Small numbers don't work with some logical operators
I discovered a wierd bug where logical comparisons with small numbers silently fail.

Example:
```
1e-20 < 1e-10
# returns TRUE

set_units(1e-20, g) < set_units(1e-10,g)
# returns FALSE
```

Here is a little script to get more details
```
triangle <- function(x) {
  for (i in 1:length(x)) {
    print(x[which(x < x[i])])
  }
}

l = 10^(c(-9:-4))
u = set_units(l, g)

triangle(l)
triangle(u)
```

It seems that in this case, when the magnitude of the number on the RHS is 1e-7 or less, the comparison fails.

I am running R 4.3.1 and units-0.8-2 on the latest Linux Mint Cinnamon.

``` 
> R.version

platform       x86_64-pc-linux-gnu         
arch           x86_64                      
os             linux-gnu                   
system         x86_64, linux-gnu           
status                                     
major          4                           
minor          3.1                         
year           2023                        
month          06                          
day            16                          
svn rev        84548                       
language       R                           
version.string R version 4.3.1 (2023-06-16)
nickname       Beagle Scouts               

```

Enchufa2:
Closes #351


Enchufa2:
@edzer I think we should make a new patch release for this one.

edzer:
Great work!

> I think we should make a new patch release for this one.

I Agree!

Enchufa2:
Great, please let me know if there's any issue with the submission.

PR #371, opened 2024-08-12
Add `cbind` and `rbind` methods

### LINKED ISSUE #311: cbind and rbind silently drop units
`cbind()` and `rbind()` drop the units when combining with numeric or other units.

``` r
x <- units::set_units(c(1, 2), "g")
cbind(1, x)
#>        x
#> [1,] 1 1
#> [2,] 1 2
rbind(1, x)
#>   [,1] [,2]
#>      1    1
#> x    1    2
cbind(x, x)
#>      x x
#> [1,] 1 1
#> [2,] 2 2
rbind(x, x)
#>   [,1] [,2]
#> x    1    2
#> x    1    2
```

jranke:
Fixes #311


Enchufa2:
Thanks. This PR currently breaks `quantities`. I'd like to avoid that unless strictly necessary. Did you look into it by any chance?

jranke:
> Thanks. This PR currently breaks `quantities`. I'd like to avoid that unless strictly necessary. Did you look into it by any chance?

I had a quick look, but it gets a bit complicated. The proposed method `cbind.units` is called in `errors:::propagate` in the case of quantities and there it fails. I do not see a nice way to avoid or fix this.

Enchufa2:
No problem. Let's put this on hold then until I find the time to have a look.

JohnADawson:
@Enchufa2, when do you expect `cbind` and `rbind` to be added?

Enchufa2:
I forgot about this PR, thanks for the reminder. I'll take a look to try to push an update to CRAN on January with this.

Enchufa2:
Ok, so I'll have to land fixes for quantities and errors first, but this is good to go. Thanks for the contribution!

PR #373, opened 2024-09-02
Adopt new way of registering methods

olivroy:
I wonder if this is still necessary?

https://github.com/r-quantities/units/blob/123c23d46a570bbe97bd3900bf332d2df9804654/R/scale_units.R#L135-L140

Looks like this was added in 2021.


Enchufa2:
I'm sorry, but could you explain the PR in the first place, please? I don't remember discussing why this is needed or better than the current method. Is there some bug that needs fixing? AFAIK works fine now.

olivroy:
Apologies for the noise. This probably doesn't change anything

PR #394, opened 2025-02-14
add passthrough for `keep_units()` when units are not provided

### LINKED ISSUE #392: `keep_units()` with non-units `x` input
If `keep_units()` is called with a non-units `x` value, would it be reasonable to just pass through that value to the normal function and not apply units at the end? For example:

``` r
units::keep_units(rexp, n = 100, x = 3)
#> Error in UseMethod("units"): no applicable method for 'units' applied to an object of class "c('double', 'numeric')"
```

<sup>Created on 2025-02-13 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This would be helpful in a function I'm writing where I don't know in advance if the user will provide an input with units or not. As a workaround I can check for units on the input and only call `keep_units()` when there are units to keep, but built-in behavior would be great!

d-morrison:
fixes #392

### DIFF
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -19,7 +19,7 @@
 #'
 #' @export
 keep_units <- function(FUN, x, ..., unit=units(x)) {
-  if (inherits(x, "units") || inherits(unit, "symbolic_units")) {
+  if (inherits(try(unit, silent = TRUE), "symbolic_units")) {
     set_units(do.call(FUN, list(x, ...)), unit, mode = "standard")
   } else {
     warning("`x` does not have units.") # maybe not necessary to warn?
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -6,6 +6,11 @@
 #' Provided for incompatible functions that do not preserve units. The user is
 #' responsible for ensuring the correctness of the output.
 #'
+#' If \code{x} is not a \code{units} object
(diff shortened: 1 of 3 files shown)


d-morrison:
@Enchufa2 I think this covers all the cases we discussed in #392? Hope it's helpful; feel free to discard if not.

R/helpers.R
```diff
@@ -19,7 +19,12 @@
 #'
 #' @export
 keep_units <- function(FUN, x, ..., unit=units(x)) {
-  set_units(do.call(FUN, list(x, ...)), unit, mode="standard")
+  if (inherits(try(unit, silent = TRUE), "symbolic_units")) {
+    set_units(do.call(FUN, list(x, ...)), unit, mode = "standard")
+  } else {
+    warning("`x` does not have units.")
```
Enchufa2, line 25: Maybe "wrong `unit` specification" or something similar? Because the user may have defined `unit` explicitly and still get it wrong.

Enchufa2 (changes requested):
Would you mind adding some details about this usage to the documentation of the function? Also maybe some examples about this random generation with days, years, rates...

Enchufa2:
@d-morrison Kind reminder. :)

Enchufa2 (approved):
Thanks!

PR #399, opened 2025-03-04
add function to convert units

### LINKED ISSUE #398: Add unit convert function?
Coming from using udunits2, I am surprised not to find an equivalent to `udunits::ud.convert()`. This is a utility that I frequently use to avoid errors in conversion. 

It might look something like:

```r
convert_units <- function(x, from, to) {
  x |> set_units(from) |>
    set_units(to) |> 
    drop_units()
}
```

But the fact that it is missing makes me wonder if it was an explicit choice not to include such a function. If not, let me know and I can submit a PR.

dlebauer:
Initial implementation of a function to perform unit conversion without converting to unit class. Fixes #398

Currently duplicates `ud_convert` defined in the auto-generated RcppExports.R, not sure where the function should live, but currently passes `devtools::check()`


Enchufa2:
We cannot just overwrite the RcppExports function. With these changes, I have

- Renamed the C++ function, so that `ud_convert` can be implemented and exported independently.
- Added support for `symbolic_units` to `ud_convert`.
- Simplified `ud_are_convertible` in a similar way.
- Merged the manual pages and move some tests around to better reflect the code structure.

Any further comments?

PR #416, opened 2025-09-28
Implement new tokenizer

### LINKED ISSUE #221: squared unit behaves differently in numerator and denominator
Thanks for a great package.

I notice that both m^2 and m2 share a common internal representation when used in a numerator, but not for a denominator.  Should behaviors match?

```
> as_units('m^2')
 1 [m^2] 
> as_units('m2') 
1 [m^2] 
> as_units('kg/m^2') 
1 [kg/m^2] 
> as_units('kg/m2') 
1 [kg/m2] 
> str(as_units('m^2')) 
Object of class units:  num 1  - attr(*, "units")=List of 2  
 ..$ numerator  : chr [1:2] "m" "m"  
 ..$ denominator: chr(0)   
 ..- attr(*, "class")= chr "symbolic_units" 
> str(as_units('m2')) 
Object of class units:  num 1  - attr(*, "units")=List of 2  
 ..$ numerator  : chr [1:2] "m" "m"  
 ..$ denominator: chr(0)   
 ..- attr(*, "class")= chr "symbolic_units" 
> str(as_units('kg/m^2')) 
Object of class units:  num 1  - attr(*, "units")=List of 2   
..$ numerator  : chr "kg"   
..$ denominator: chr [1:2] "m" "m"   
..- attr(*, "class")= chr "symbolic_units" 
> str(as_units('kg/m2')) 
Object of class units:  num 1  - attr(*, "units")=List of 2   
..$ numerator  : chr "kg"   
..$ denominator: chr "m2"  
 ..- attr(*, "class")= chr "symbolic_units"

```

### LINKED ISSUE #383: Wrong parsing of numeric values as part of units
It seems like strings representing integers cannot be coerced to units (except for 1) and "dot" means exponentiation (except after zero).  Can you point me to the documentation?  Thanks in advance.

```
as_units('1')
1 [1]
> as_units('2')
Error in as_units.call(expr, check_is_valid = check_is_valid) : 
  is.language(x) is not TRUE
> as_units('2.1')
1 [2.]
> as_units('2.2')
1 [2.^2]
> as_units('2.3')
1 [2.^3]
> as_units('0.3')
Error: In ‘`0.`^(3)’, ‘0.’ is not recognized by udunits.
```

Enchufa2:
Closes #221, closes #383.

I would like to request feedback from everyone involved:

- @bergsmat, @pepijn-devries, @gitrdm as the ones affected by the previous behavior;
- @edzer, @t-kalinowski as the authors of the previous mechanism.

If you have time, it would be great if you could test this, maybe discover issues I didn't consider.

This is a complete rework of the unit construction code. Previously, unit strings were adapted with a bunch of regex so that they could be parsed into R expressions, which then would be evaluated in an environment populated with the individual units. In fact, R expressions are much more general than what's needed here, so I've decided to go in the other direction. Instead, expressions are converted to strings, and a much simpler and faster C++ tokenizer gets the job done, no regex required.

TL;DR, this tokenizer expects multiplications and divisions of numbers and unit names/symbols, with parentheses and integer exponents. Both numbers and symbols are treated as individual tokens, so that things like `ml/min/1.73/m^2` work.


pepijn-devries:
Hi @Enchufa2 ,

The proposed changes are in the `tokenizer` branch right? I have some specific tests for unit conversions in my package, so I will test how it behaves with these modifications. I will report back here...

pepijn-devries:
In general it sound like a smart strategy to keep it simple and fast. In most cases the new behaviour works just as before. However, there are some exotic cases where corrections are required.

Consider the string `"oz/10 gal/1000 ft2"`. In the previous release of `units` this was translated as `"oz/1000ft2/10gal"` which is ok. The new implementation returns `"oz*gal*ft^2/10/1000"` which is absolutely not correct. Could we expect the `units` package to sanitise the input string to fix this, or should the user do this?

Anyway after adding explicit brackets in the string this was fixed again:
`units::as_units("oz/(10gal)/(1000ft2)")` returns `[oz/10/gal/1000/ft^2]`

Please let me know when/if you plan to make this the definitive new behaviour. It will require more thorough sanitation on my behalf

Enchufa2:
Currently, the tokenizer switches to the denominator when a `/` is found, and switches back to the numerator when a token is read, number or symbol. This is strictly correct mathematically speaking, and this is what udunits2 does, as @gitrdm points out. Instead, your interpretation is that a numerical value goes somewhat _attached_ to the next symbol.

I don't have a strong opinion here. We could support this interpretation by switching only after a symbol, but not after a number. Switching interpretation could be an option of the package, and if this is common enough, we could make it the default.

Enchufa2:
@pepijn-devries With the latest commit, strict tokenization can be enabled via an option, but it's off by default, meaning that numbers are effectively treated like prefixes:

``` r
units::as_units("oz/10 gal/1000 ft2")
#> 1 [oz/10/gal/1000/ft^2]
units::as_units("ml/min/1.73m^2")
#> 1 [ml/min/1.73/m^2]
```

Now the question is whether this non-strict _numbers-as-prefixes_ mode should be propagated to the formatting, and therefore whether we should remove the `/` between number and unit. What do you think?

Enchufa2:
@alwinw Adding you to this conversation because I found your package {epocakir} deals with `ml/min/1.73m^2` measurements.

Enchufa2:
Greater than the one currently on CRAN. :) Given the importance of the change, I will make it v1.0.

Enchufa2:
I've run a full revdep check and there's only one package failing with this PR ({epocakir} @alwinw, see [the report](https://github.com/r-quantities/units/blob/1c81303825e8b89199fad62e5028e2e9a97cb5bc/revdep/problems.md)), which is very good news.

Conjuring up @billdenney too in case you have some time to test this.

billdenney:
I doubt that I will have time to test it soon. My challenges often come with math done on units. For example, will this handle something like `log(set_units(1, "kg/m^2"))`?

Enchufa2:
I'm happy with the result. I'm planning to merge this by the end of the week. I'll be happy to receive further feedback or otherwise I'll interpret administrative silence as agreement. ;-)

alwinw:
> I've run a full revdep check and there's only one package failing with this PR ({epocakir} @alwinw, see [the report](https://github.com/r-quantities/units/blob/1c81303825e8b89199fad62e5028e2e9a97cb5bc/revdep/problems.md)), which is very good news.
> 
> Conjuring up @billdenney too in case you have some time to test this.

Thanks for the heads up, I'll make an update for my package thanks!

Issue #179, opened 2018-11-24
Power when units are dimensionless


jmerc13:
Hi,
First, thanks for this excellent package! I'm hoping to build on top of it (next to it?) for some physically-based modeling in R.

Second, I have a suggestion for implementing power function in some cases:
```
library(units)
x <- as_units(1:12, "kPa")
y <- as_units(seq(0.1, 1.2, by = 0.1), "km")
z <- x / y #units: [kPa/km]
k <- y / y #units: [1]; unitless

#Test case 1: y/y, generates a unitless product, but can't be used in power.
z ^ (2 * y / y) 
#Error in Ops.units(z, (2 * k)) : 
#  power operation only allowed with numeric power

#Test case 2: same as Test case 1, but initial calculation of the power value is outside of the power function.
z ^ (2 * k)
#Error in Ops.units(z, (2 * k)) : 
#  power operation only allowed with numeric power
```
I think both of the above errors are forced, rather than actual, because once a unit becomes dimensionless (i.e., [1]) it should act as any other dimensionless numeric or integer. I think an additional if statement for this condition may therefore be needed, to allow the above test cases.

I'm still pretty new to GitHub, but I could try to implement this, if you'd like.

Thanks!

Enchufa2:
It's not so simple, because, for example, a radian is a unit that _is not a unit_: it's dimensionless, but at the same time, I think we shouldn't be able to use radians as exponents.

So the current approach is to limit what you can do with _any_ unit, dimensionless or not, and thus, the user is responsible for dropping the unit if that's appropriate. I don't think that taking the slippery path of making exceptions would be a good design decision, especially when they add no new functionality.

jmerc13:
Good points. But it seems to me that dimensionless units like radians are more the exception than the rule. So my instinct would be to organize around the more common case, and not the corner case. Given these thoughts, I kind of wonder if there aren't two kinds of dimensionless, in this case: 1) truly dimensionless values (such that units have cancelled out), and 3) quasi-dimensionless, such as the case of radians.

jmerc13:
Ha! Good point. I forgot that's how rads worked. But I still think that is not very common to generate a unit when the underlying units (e.g., `L/L`) cancel. So I still think it might be worth revisiting this issue.

edzer:
The only thing that makes `rad` special, IMO, is that it has a name; otherwise it is just another unitless. There are more named unitless, but most don't have a name.

[...]

PR #435, opened 2026-06-23
simplify unitless powers automatically

### LINKED ISSUE #434: Use power operations with dimensionless units
This is in fact a revivial of #179 with a reprex, calculating the decline of a concentration of 10 µg/L with a decline rate constant of one per hour over three hours.

``` r
library(units)

c0 <- set_units(10, "µg/L")

k_1 <- set_units(24, "1/d")
k_2 <- set_units(1, "1/h")
k_1 == k_2
#> [1] TRUE

t <- set_units(3, "h")

c0 * exp(-k1 * t)
#> Error:
#> ! object 'k1' not found

c0 * exp(drop_units(-k_1 * t))
#> 0.4978707 [µg/L]
c0 * exp(drop_units(-k_2 * t))
#> 0.4978707 [µg/L]
```

<sup>Created on 2026-06-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I believe the call to `drop_units` should not be necessary. But as we can see, units are reconciled before they are dropped, so it is not a big issue.

Enchufa2:
Fixes #434, revisiting #179.

@edzer This is my proposal with no impact on other existing functionality. What do you think?

### DIFF
--- a/R/arith.R
+++ b/R/arith.R
@@ -129,7 +129,7 @@ Ops.units <- function(e1, e2) {
 
   } else if (pw) {
     if (e2_inherits_units) {
-      if (e1_inherits_units && identical(units(e1), units(as_units(1)))) {
+      if (e1_inherits_units && identical(units(e1), unitless)) {
         e1 <- drop_units(e1)
       } else if (e1_inherits_units) {
         stop("power operation only allowed with numeric power")
@@ -179,7 +179,7 @@ Ops.units <- function(e1, e2) {
     # when the power is negative and we have a special case when it is zero where
     # units should be removed.
     if (e2 == 0) {
-      u <- units(as_units(1))
+      u <- unitless
     } else {
       tbl_den <- tabulate(factor(units(e1)$denominator))
       tbl_num <- tabulate(factor(units(e1)$numerator))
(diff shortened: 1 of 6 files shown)


R/arith.R
```diff
@@ -128,12 +128,14 @@ Ops.units <- function(e1, e2) {
...
+    if (e2_inherits_units && identical(units(e2), unitless)) {
+      e2 <- drop_units(e2)
+      if (!e1_inherits_units)
+        return(NextMethod())
+
+    } else if (e2_inherits_units) {
+      if (e1_inherits_units && identical(units(e1), unitless))
         e1 <- drop_units(e1)
-      } else if (e1_inherits_units) {
-        stop("power operation only allowed with numeric power")
```
Enchufa2, line 135: Not needed. "wrong base in power operation" is triggered later anyways.

jranke:
LGTM, thanks for addressing this so quickly! Impressive!

R/arith.R
edzer: LGTM!

edzer:
Sure, LGTM!

